PR #46, opened 2016-12-12
Integration - Adapt Forest Liana integration to new secrets names

arnaudbesnier:


services/deploy/heroku.js
```diff
@@ -66,13 +66,13 @@ function Heroku(config) {
       });
   };
 
-  this.updateForestSecretKey = (appName, secretKey) => {
+  this.updateForestEnvSecret = (appName, envSecret) => {
```
arnaudbesnier: Where is the method that set the `FOREST_AUTH_SECRET` while deploying on heroku?

PR #131, opened 2018-05-04
Support password with special characters

root-io:


root-io:
[(?) Database - Connection ensure special characters for password work on admin database connection (# for example)](https://trello.com/c/tLZ6oi6U/2498-database-connection-ensure-special-characters-for-password-work-on-admin-database-connection-for-example)

PR #134, opened 2018-05-09
Fix heroku deploy with password that contain special characters

root-io:


services/dumper.js
```diff
@@ -1,4 +1,5 @@
 'use strict';
+const { getDatabaseUrl } = require('utils');
```
arnaudbesnier: As a convention, we try to require our own components after external dependencies (bluebird, fs, lodash,... ).
I feel it makes the code more readable.

SeyZ:
We've decided to remove the heroku-deploy command.

PR #141, opened 2018-06-22
Fix empty schema error log

lauriejim:
Update `lumber update` error log when database is empty:

`💀  Oops, your database is empty. Please, create some tables before running Lumber update.💀`

instead of

`💀  Oops, your database is empty. Please, create some tables before running Lumber generate.💀`

`generate` => `update`


SeyZ:
Thanks @lauriejim !

Issue #52, opened 2016-12-28
No support for MySQL enum and float field type?


virtser:
Following #51, I found that MySQL enum and floar fields are simply ignored by Forest and not displayed at all.

We have the following field in our DB model which works and displays just fine using Sequelize:

```
    status: {
      type: DataTypes.ENUM,
      values: ['pending', 'listed'],
      allowNull: false,
      defaultValue: 'pending'
    },
    display_order: {
      type: DataTypes.FLOAT
    }
```

virtser:
I found that lumber creates its own models using sequelize, but it doesn't detect well the column data types. Changing it manually solves the issue.

SeyZ:
Hey @virtser ! Thanks for report! If you have some time, can you help Lumber supporting your data types natively by patching the following method according to your schema: https://github.com/ForestAdmin/lumber/blob/master/services/table-analyzer.js#L29

Thank you!

arnaudbesnier:
Manual changes are a workaround but definitly not a fix.
This issue should be reopened.

ArnaudValensi:
This is fixed in the last release (1.10.0)

jeremymaignan:
Hi, 
I am facing the same issue. How can I fix it ? 
Thanks

PR #145, opened 2018-06-27
 Detect enum type for model generation 

ArnaudValensi:


ArnaudValensi:
[(?) P3 - Search on ENUM - "SequelizeDatabaseError: function lower("enum_Table_status") does not exist"](https://trello.com/c/nmY4Hwa3/2649-p3-search-on-enum-sequelizedatabaseerror-function-lowerenumtablestatus-does-not-exist)

services/table-analyzer.js
```diff
@@ -23,26 +22,53 @@ function TableAnalyzer(queryInterface, config) {
...
+      case 'NVARCHAR': // MSSQL type
         return 'STRING';
+      case 'USER-DEFINED': {
+        const isEnum = await checkIfColumnIsTypeEnum(columnName);
+
+        return isEnum ? `ENUM('${special.join('\', \'')}')` : 'STRING';
+      }
+      case (type.match(mysqlEnumRegex) || {}).input:
+        return type;
       case 'UNIQUEIDENTIFIER':
```
ArnaudValensi: There is no enum in mssql. A common way the achieve the same result (and it is what sequelize does) is to create a varchar and create a "check" contraint to only accept certain values.
So here we don't care, since it will be seen as a `STRING`

services/table-analyzer.js
```diff
@@ -23,26 +22,53 @@ function TableAnalyzer(queryInterface, config) {
...
       case (type.match(/TEXT.*/i) || {}).input:
       case (type.match(/VARCHAR.*/i) || {}).input:
       case (type.match(/CHAR.*/i) || {}).input:
-      case 'NVARCHAR': // MSSQL type
+      case 'NVARCHAR': // MSSQL type
         return 'STRING';
+      case 'USER-DEFINED': {
+        const isEnum = await checkIfColumnIsTypeEnum(columnName);
+
+        return isEnum ? `ENUM('${special.join('\', \'')}')` : 'STRING';
```
arnaudbesnier: Isn't non empty `special` + type `USER-DEFINED` enough to detect that the column is an ENUM ?

services/table-analyzer.js
```diff
@@ -23,26 +22,53 @@ function TableAnalyzer(queryInterface, config) {
       case 'sqlite':
         query = `PRAGMA foreign_key_list('${table}');`;
         break;
+      default:
+        break;
```
SeyZ: This is not the default behavior?

services/table-analyzer.js
```diff
@@ -23,26 +22,53 @@ function TableAnalyzer(queryInterface, config) {
...
+  function checkIfColumnIsTypeEnum(columnName) {
+    const query = `
+      SELECT i.udt_name
+      FROM pg_catalog.pg_type t
+      JOIN pg_catalog.pg_namespace n ON n.oid = t.typnamespace
+      JOIN pg_catalog.pg_enum e ON t.oid = e.enumtypid
+      JOIN information_schema.columns i ON t.typname = i.udt_name
+      WHERE i.column_name = '${columnName}'
+      GROUP BY i.udt_name;
+    `;
```
SeyZ: This function seems to only works for postgres? There's no context using `queryInterface.sequelize.options.dialect`. Is it normal?

ArnaudValensi: Done

SeyZ: Done? Don't see the change?

ArnaudValensi: The check is done before, since we only want to check if it is an enum when we are using postgres. https://github.com/ForestAdmin/lumber/pull/145/files#diff-3f431298269ee4103c03ca56c7f393a4R66

arnaudbesnier:
@ArnaudValensi it might also fix a part of this issue https://github.com/ForestAdmin/lumber/issues/52.
Maybe you should add a comment in this issue if you think there is an improvement with your code.

package.json
```diff
@@ -31,8 +32,10 @@
     "winston": "^2.2.0"
   },
   "devDependencies": {
+    "ajv": "^6.0.1",
     "ava": "0.25.0",
     "chai": "4.1.2",
+    "eslint": "^4.9.0",
```
arnaudbesnier: Those are dev dependencies, thus, not so critical, but I would definitely hardcode the versions.

PR #172, opened 2018-10-03
On-boarding - Improve the on-boarding experience

SeyZ:


SeyZ:
[Users want to have a better Lumber on-boarding](https://trello.com/c/94YTnUPn/2938-users-want-to-have-a-better-lumber-on-boarding)

services/authenticator.js
```diff
@@ -61,12 +62,17 @@ function Authenticator() {
...
+    let rendering;
 
     return agent
       .post(`${config.serverHost}/api/guests`)
       .set('forest-origin', 'Lumber')
       .send(new GuestSerializer(guest))
       .then((response) => new GuestDeserializer.deserialize(response.body))
+      .then((guest) => {
+        rendering = guest.project.defaultEnvironment.renderings[0];
+        return guest;
```
VincentMolinie: `return { guest, rendering }` and remove support `let rendering`

services/authenticator.js
```diff
@@ -61,12 +62,17 @@ function Authenticator() {
...
+      })
       .then((guest) => {
```
VincentMolinie: here just do `({ guest, rendering })`

services/authenticator.js
```diff
@@ -111,7 +117,11 @@ function Authenticator() {
               .then((environment) => {
                 project.defaultEnvironment = environment;
                 return project;
-              });
+              })
+              .then((project) => {
+                project.defaultEnvironment.renderings = [rendering];
+                return project;
+              })
```
VincentMolinie: missing semi-colon

services/authenticator.js
```diff
@@ -61,12 +62,17 @@ function Authenticator() {
 
   this.registerAndCreateProject = (config) => {
     let guest = { email: config.email };
```
VincentMolinie: `const guestParams`, it's confusing with the one after

services/authenticator.js
```diff
@@ -61,12 +62,17 @@ function Authenticator() {
...
+    let rendering;
 
     return agent
       .post(`${config.serverHost}/api/guests`)
       .set('forest-origin', 'Lumber')
       .send(new GuestSerializer(guest))
       .then((response) => new GuestDeserializer.deserialize(response.body))
```
VincentMolinie: remove parentheses around parameter

VincentMolinie: `.then(response => new GuestDeserializer.deserialize(response.body))`

services/authenticator.js
```diff
@@ -61,12 +62,17 @@ function Authenticator() {
...
+      .then((guest) => {
+        rendering = guest.project.defaultEnvironment.renderings[0];
+        return guest;
+      })
       .then((guest) => {
         let user = {
```
VincentMolinie: `const user`

services/authenticator.js
```diff
@@ -61,12 +62,17 @@ function Authenticator() {
...
           id: guest.user.id,
```
VincentMolinie: `const userPayload`

services/authenticator.js
```diff
@@ -111,7 +117,11 @@ function Authenticator() {
               .then((environment) => {
                 project.defaultEnvironment = environment;
```
VincentMolinie: why not setting your rendering here ?

README.md
```diff
@@ -44,11 +48,13 @@ Full demo video: https://www.youtube.com/watch?v=2cKSsBxrvR8
 
 `$ lumber [command]`
 
-- `generate`      generate your admin interface based on your database schema
+- `generate`      generate your back office based on your database schema
 - `update`        update your models's definition according to your database schema
+- `deploy`        deploy your back office application to production
 - `user`          show your current logged user
 - `login`         sign in to your Forest account
```
VincentMolinie: I would say _`Log in to your Forest account`_ than sign in. Most of the people do not understand the difference between sign in and sign up. Just an advise

services/authenticator.js
```diff
@@ -140,6 +151,32 @@ function Authenticator() {
       });
   };
 
+  this.logout = async (opts) => {
```
VincentMolinie: This should be call in logout to avoid duplicated code

test/sample-project/app.json
```diff
@@ -1,28 +0,0 @@
-{
```
VincentMolinie: Why removing this ? What was it's purpose ?

templates/app/env
```diff
@@ -4,5 +4,6 @@ FOREST_ENV_SECRET=<%= forestEnvSecret %>
 <% if (!ssl) {%>#<% } %>SSL_DATABASE=true
 <% if (!encrypt) {%>#<% } %>ENCRYPT_DATABASE=true
 <% if (forestUrl) {%>FOREST_URL=<%= forestUrl %><% } %>
+<% if (devRenderingId) {%>DEV_RENDERING_ID=<%= devRenderingId %><% } %>
```
VincentMolinie: `DEV_RENDERING_ID` ? It's not really explicit what is its purpose, `REDIRECT_DEFAULT_RENDERING_ID` ?

templates/app/env
```diff
@@ -4,5 +4,6 @@ FOREST_ENV_SECRET=<%= forestEnvSecret %>
...
 <% if (dbSchema) {%>DATABASE_SCHEMA=<%= dbSchema %><% } %>
-<% if (port) {%>PORT=<%= port %><% } %>
\ No newline at end of file
+<% if (port) {%>PORT=<%= port %><% } %>
```
VincentMolinie: We could add FOREST_URL here in case we detect the admin server to be localhost to make dev faster, just a thought

services/authenticator.js
SeyZ: I just changed the `var` to `const` without touching the rest of the code.
Your suggestion is great but I'm afraid it will change a lot of codes and should be on another scheduled change IMO

templates/app/env
SeyZ: I don't agree. The variable is really "the development rendering id" and the use case today is for redirection but it could be used later for another purpose.

templates/app/env
SeyZ: Let's do this on another PR I think

README.md
SeyZ: Will be on another PR with a complete review on the help + readme

test/sample-project/app.json
SeyZ: It was a file used to deploy on heroku automatically. Today this command disappeared.
This file was generated with an old version of Lumber in the `test/sample-project` folder.
So it's useless.

VincentMolinie: Okey good to know

lumber-logout.js
```diff
@@ -3,18 +3,15 @@ const os = require('os');
...
-    fs.unlinkSync(path);
-    console.log(chalk.green('👍  You\'re now unlogged 👍 '));
-  } else if (err.code === 'ENOENT') {
-    logger.error('🔥  You\'re not logged 🔥');
-  }
-});
+(async () => {
+  await authenticator.logout({ log: true })
+    .catch((err) => {
+      console.error(err);
```
VincentMolinie: I'm not sure catching the error here is usefull. The error will still be printed if not catch

services/authenticator.js
```diff
@@ -140,6 +147,32 @@ function Authenticator() {
...
+
+          if (opts.log) {
+            console.log(chalk.green('👍  You\'re now unlogged 👍 '));
+          }
+
+          resolve();
+        } else if (err.code === 'ENOENT') {
+          if (opts.log) {
+            logger.error('🔥  You\'re not logged 🔥');
+          }
```
VincentMolinie: Why do we have sometimes logger and sometimes console, it's confusing ?

lumber-logout.js
```diff
@@ -3,18 +3,15 @@ const os = require('os');
 const fs = require('fs');
 const chalk = require('chalk');
```
VincentMolinie: Remove the old required

services/authenticator.js
SeyZ: `console.error is used to print the full stack trace`

VincentMolinie: Yeah I know but what I mean is why don't we migrate it to logger ? And why don't we use `logger.success` for example to print a success message instead of using `console.log` with `charn`

PR #179, opened 2018-11-05
[*] Technical - Let Forest developers override the default url for the client app

arnaudbesnier:


SeyZ:
In my opinion, you should not write any logic in the generated back offices for all the Lumber's users.

services/dumper.js
```diff
@@ -130,7 +130,8 @@ function Dumper(project, config) {
       ssl: config.ssl,
       encrypt: config.ssl && config.dbDialect === 'mssql',
       dbSchema: config.dbSchema,
-      port: config.appPort
+      port: config.appPort,
+      frontendHost: config.frontendHost
```
SeyZ: See my global remark on this. For me, we should keep this file untouched.

services/prompter.js
```diff
@@ -19,6 +19,10 @@ async function Prompter(program, requests) {
     envConfig.serverHost = 'https://api.forestadmin.com';
   }
 
+  if (process.env.FRONTEND_HOST) {
+    envConfig.frontendHost = process.env.FRONTEND_HOST;
+  }
+
```
SeyZ: Same here. We should keep this file untouched.

templates/app/app.js
```diff
@@ -19,7 +19,7 @@ app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'public')));
 
 app.use(cors({
-  allowedOrigins: ['*.forestadmin.com'],
+  allowedOrigins: ['*.forestadmin.com', 'localhost:4200'],
```
SeyZ: Something like: `allowedOrigins: [process.env.FOREST_HOSTNAME || '*.forestadmin,.com'],` ?

templates/app/env
```diff
@@ -7,3 +7,4 @@ FOREST_ENV_SECRET=<%= forestEnvSecret %>
 <% if (devRenderingId) {%>DEV_RENDERING_ID=<%= devRenderingId %><% } %>
 <% if (dbSchema) {%>DATABASE_SCHEMA=<%= dbSchema %><% } %>
 <% if (port) {%>PORT=<%= port %><% } %>
+<% if (frontendHost) {%>FRONTEND_HOST=<%= frontendHost %><% } %>
```
SeyZ: Same here. I prefer to set the environment variable in the dotenv manually or run my lumber back office application with: `FRONTEND_HOST=http://localhost:4200 npm start` if you don't want to set it into the .env

arnaudbesnier:
Did I set you as a reviewer? It was just a draft to help me test/debug onboardings fastly.
If I were you, I would not lose time to review something that is not clear or finished.

That being said, we should find a "dev mode" for Lumber, it might be with a single option instead of multiple configuration variables.
Without it, the dev team won't be performant to test, debug or implement features.
(I did at least 20 generates command and it was painful to setup manually the variables in the generated projects).

PR #194, opened 2019-02-07
[*] Command Update - Add the new FOREST_DB_DIALECT environment variable to be able to configure project fastly

VincentMolinie:


services/prompter.js
```diff
@@ -176,7 +180,7 @@ async function Prompter(program, requests) {
 
     if (isRequested('ssl')) {
       if (process.env.FOREST_DB_SSL) {
-        envConfig.ssl = process.env.FOREST_DB_SSL;
+        envConfig.ssl = JSON.parse(process.env.FOREST_DB_SSL);
```
VincentMolinie: This allow to do FOREST_DB_SSL=false, else it was always true

PR #198, opened 2019-03-15
[+] Docker - Add Docker build files

root-io:


root-io:
[Lumber - Add Docker build files](https://trello.com/c/DG0RJnrg/3291-lumber-add-docker-build-files)

templates/app/docker-compose.yaml
```diff
@@ -0,0 +1,9 @@
+version: '3'
+services:
+  app:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: <%= appName %>
+    ports:
+      - "<%= port %>:<%= port %>"
```
ArnaudValensi: To be more standard and avoid the user to know how forest handle environment variable, I would not copy the .env file and configure the variable directly in the docker-compose.yaml.

services/dumper.js
```diff
@@ -174,6 +174,44 @@ function Dumper(project, config) {
...
+      databaseUrl: getDatabaseUrl(),
+      forestUrl: process.env.FOREST_URL,
+      devRenderingId: project.defaultEnvironment.renderings[0].id,
+      ssl: config.ssl,
+      encrypt: config.ssl && config.dbDialect === 'mssql',
+      dbSchema: config.dbSchema,
+      port: config.appPort,
+    };
+
+    fs.writeFileSync(`${pathDest}/docker-compose.yaml`, template(settings));
```
SeyZ: It's `docker-compose.yml`, not `yaml`

templates/app/docker-compose.yaml
SeyZ: Me too!

templates/app/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+COPY package*.json ./
+
+RUN npm install
+
+COPY . .
+
+EXPOSE <%= port %>
+
+CMD ["npm", "start"]
```
SeyZ: Can we remove the empty lines between each line?

services/dumper.js
root-io: From the doc it seems we can use both:

https://docs.docker.com/compose/compose-file/#service-configuration-reference

`Tip: You can use either a .yml or .yaml extension for this file. They both work.`

templates/app/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      context: .
+      dockerfile: Dockerfile
+    container_name: <%= appName %>
+    environment:
+      - DATABASE_URL=<%= databaseUrl %>
+      - FOREST_AUTH_SECRET=<%= forestAuthSecret %>
+      - FOREST_ENV_SECRET=<%= forestEnvSecret %>
+      <% if (!ssl) {%># <% } %>- SSL_DATABASE=true
+      <% if (!encrypt) {%># <% } %>- ENCRYPT_DATABASE=true
+      <% if (forestUrl) {%>- FOREST_URL=<%= forestUrl %><% } %>
```
ArnaudValensi: We can remove this.

templates/app/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      <% if (devRenderingId) {%>- DEV_RENDERING_ID=<%= devRenderingId %><% } %>
```
ArnaudValensi: We can remove this too.

templates/app/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      <% if (dbSchema) {%>- DATABASE_SCHEMA=<%= dbSchema %><% } %>
```
ArnaudValensi: Keep this but remove the condition.

templates/app/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+      <% if (port) {%>- PORT=<%= port %><% } %>
```
ArnaudValensi: We can remove this too.

PR #199, opened 2019-03-18
[+] Technical - Warn users if an update of Lumber is available

root-io:


root-io:
[Lumber/Toolbelt - Warn user if an update is available.](https://trello.com/c/kHzTVkta/3294-lumber-toolbelt-warn-user-if-an-update-is-available)

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Technical - Warn users if an update of Lumber is available.
```
arnaudbesnier: This is not something "Technical" it is a feature.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
...
+
```
arnaudbesnier: It seems that you've done other stuff here.

package.json
```diff
@@ -13,25 +13,26 @@
...
+    "pg": "6.4.2",
     "sequelize": "git://github.com/ForestAdmin/sequelize.git#feature/support-schema",
-    "superagent": "^3.8.3",
-    "superagent-promise": "^1.1.0",
-    "tedious": "^1.14.0",
-    "winston": "^2.2.0"
+    "superagent": "3.8.3",
+    "superagent-promise": "1.1.0",
+    "tedious": "1.15.0",
+    "winston": "2.4.3"
```
arnaudbesnier: Are you confident with all those changes?

root-io: I looked at the existing package-lock and took the installed versions out of it, quite confident, I think it's worth the risk, the sooner the better

root-io:
I think that's because the packages versions are set with `^`

SeyZ:
It's a great nice to have but honestly, I'm not a big fan to have a subprocess ran to check the version. It means when something goes wrong for any reason (process.exit(1)), the subprocess is still running. In addition, there're some issue related to the subprocess: https://github.com/yeoman/update-notifier/issues/140.

I prefer to close this PR at the moment.

PR #200, opened 2019-03-25
Lumber reads .env vars

josh-egan:
While attempting to use `lumber generate --connection-url`, I discovered that the program did not correctly read in the `.env` file. While this may not be a typical use case (I suspect that the most common use case is to run the generate command from a generic directory without a .env file), I have been running the `generate` command repeatedly as our project has evolved, and it would be really useful to have the code in the `Prompter` class work, which is already capable of pulling the connection string from the env.

This change also appears to have fixed an issue I was seeing with `lumber update --connection-url`. Prior to this change, I was still getting prompted to input the connection url in the terminal. However, with this change in place, the environment variable is getting read correctly and used by the program as expected. I have not taken the time to uncover why it wasn't working before and subsequently why it is now working, but it is working as expected now with these changes applied, so that's good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Improvement - Reduce the number of http requests to init the project.
 - Technical - Rename environment variable SERVER_HOST to FOREST_URL.
 
+### Fixed
+- Command Update - Environment variables are now read in from the .env file if the file exists.
+
 ## Release 1.14.1
 ### Fixed
 - Command Update - Add the 'ssl' prompt option.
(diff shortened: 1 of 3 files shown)


arnaudbesnier:
Hi @josh-egan, I created an issue about the `--connection-url` option: https://github.com/ForestAdmin/lumber/issues/209

I think that if you could do a command like this:
`lumber generate --connection-url postgres://forest:secret@localhost:5435/forest`
you would not need any specific environment variables.

What do you think?

SeyZ:
I take your feedback into account and implement the following behavior in the devel branch:
`lumber generate -c postgres://forest:secret@localhost:5446/mydatabase mysuperadminpanel`

PR #201, opened 2019-03-25
Fix prompter

josh-egan:
The prompter was skipping all subsequent prompts when the --connection-url
option was used. This leads to the generate command failing because the necessary
information is not collected. This PR fixes the prompter so that only
the prompts for the connection url pieces are skipped when the
--connection-url option is used.

The PR also fixes the reading in of `process.env.FOREST_DB_SSL`. It was previously left as a string, which evaluates to true for any valid string. This PR fixes that by parsing the value into a boolean.

### DIFF
--- a/services/prompter.js
+++ b/services/prompter.js
@@ -74,24 +74,6 @@ async function Prompter(program, requests) {
       }
     }
 
-    if (isRequested('dbSchema')) {
-      if (process.env.DATABASE_SCHEMA) {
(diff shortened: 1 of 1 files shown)


SeyZ:
Thanks for the contribution. I take inspiration from your PR to make a big clean-up of our prompter, to fix the connection-url argument and to parse the ssl value correctly.

PR #203, opened 2019-03-25
Regenerate models

josh-egan:
This PR builds on #202, so this PR should not be merged in until after #202 has been reviewed and merged.

This PR adds a new function, `regenerateModels`, which can be used by an implementing node application to blow away the existing models folder and completely regenerate the model files.

In order to accomplish this, refactoring of the `Dumper` class and its usages had to be done first. With the refactorings in place, which should not change the existing behavior, the `regenerateModels` function could then be written.

The changes for this PR, ignoring the previous commits that are part of the previous PRs, can be viewed with this link: https://github.com/ForestAdmin/lumber/compare/b4f5756...11130f7


SeyZ:
Hey @josh-egan I'm not sure about the utility of the regenerateModels function. Most of the time, Lumber users scaffold their app with Lumber and then enrich models with many things (more relationships, validations, etc.).

Maybe it will be easier for me to understand what problem you're trying to solve first?

josh-egan:
For our use case, we are looking into using Forest as an admin UI. We already have a domain layer where our models are defined. Our model definitions are not compatible with what Forest wants, and we don't want to maintain a second set of models just for Forest. Being able to regenerate the models programatically would allow us to trigger a deploy of our forest api, which would include re-generating the models that the forest api requires, whenever our domain layer is deployed.

SeyZ:
Lumber is a scaffolding tool and I'm still not sure it's a good idea to keep the default generated models without any extension (e.g. hasMany relations, validations, etc.).

The real problem is that you don't want to maintain a second set a models for your admin panel, which I completely understand. But I'm not sure the solution fixes this problem in an elegant way.

If you still want to do this, why not just simply write a sh script like `rm -fr models/ && lumber update` to force regenerating your models?

PR #208, opened 2019-04-02
[-] Improvement - Reduce the number of http requests to init the project (to discuss)

arnaudbesnier:


templates/app/bin/www
```diff
@@ -47,15 +47,6 @@ function onListening() {
...
   console.log(chalk.cyan(`🌳  Access the UI: http://app.forestadmin.com  🌳`));
-
-  try {
-    if (process.env.DEV_RENDERING_ID) {
-      opn(`http://app.forestadmin.com/${process.env.DEV_RENDERING_ID}`);
-    }
-  } catch (err) {
-    // NOTICE: This err is not a big deal. For example, it cannot work on
-    // production.
-  }
```
SeyZ: Ok to remove `opn` but we should also get rid of the DEV_RENDERING_ID env variable right?

SeyZ: Same for the require on top of this file and the `opn` package from the package.json, right ?

arnaudbesnier: FYI, I just recreated the initial PR made by Vince not to lose his work after having rolled back the code on the `master` branch.
The major issue with the PR in the current state is that you need to be authenticated to do a `generate`, and thus cannot create a user from the CLI anymore.

PR #236, opened 2019-07-30
Add support for custom database schema

kenanpulak:
* This pull request adds support for a custom schema for the install and update lumber scripts


lumber-update.js
```diff
@@ -42,6 +42,8 @@ program
     process.exit(1);
   }
 
+  config.dbSchema = process.env.DATABASE_SCHEMA
```
VincentMolinie:
```suggestion
  config.dbSchema = process.env.DATABASE_SCHEMA;
```

PR #240, opened 2019-08-07
[-] Command Generate - Normalize the fields name on generate inside models

VincentMolinie:


VincentMolinie:
[P3 - Snake Case DB - When using snaked_case in db forest-express-sequelize doesn't handle well columnName](https://trello.com/c/PPGCVuIG/3571-p3-snake-case-db-when-using-snakedcase-in-db-forest-express-sequelize-doesnt-handle-well-columnname)

PR #242, opened 2019-08-07
[-] Command Generate - Change model name on generate to have more conventionnal naming

VincentMolinie:


services/database-analyzer.js
```diff
@@ -17,6 +20,11 @@ function DatabaseAnalyzer(databaseConnection, config) {
     return Object.keys(schema).filter(column => schema[column].primaryKey);
   }
 
+  function formatModelName(table) {
+    const modelName = _.camelCase(_.singularize(table));
+    return `${modelName.charAt(0).toUpperCase()}${modelName.slice(1)}`;
```
arnaudbesnier: - Singularize ok
- Camelize VS Capitalize 🤔
Why did you say that first capital letter is more conventional? I am not so sure as I don't see that in Sequelize documentation, and our own backend uses camelized names for models.

VincentMolinie: In the docs yes it is: https://sequelize.org/master/manual/models-definition.html

PR #245, opened 2019-08-08
[-] Technical - Factorize duplicated code

arnaudbesnier:


services/database-analyzer.js
```diff
@@ -84,7 +116,12 @@ function DatabaseAnalyzer(databaseConnection, config) {
       if (typeof table === 'object') { table = table.tableName; }
 
       const analysis = await analyzeTable(table);
-      schema[table] = { fields: analysis[0], references: analysis[1], primaryKeys: analysis[2] };
+      schema[table] = {
```
VincentMolinie: Why just not returning this json in analyzeTable, it's way cleaner

services/database-analyzer.js
```diff
@@ -64,7 +91,12 @@ function DatabaseAnalyzer(databaseConnection, config) {
           }
         });
 
-        return [fields, references, primaryKeys];
+        const options = {
+          underscore: isUnderscored(fields),
+          timestamps: hasTimestamps(fields),
+        };
+
+        return [fields, references, primaryKeys, options];
```
VincentMolinie:
```suggestion
        return { fields, references, primaryKeys, options };
```

services/database-analyzer.js
```diff
@@ -64,7 +91,12 @@ function DatabaseAnalyzer(databaseConnection, config) {
           }
         });
 
-        return [fields, references, primaryKeys];
+        const options = {
+          underscore: isUnderscored(fields),
+          timestamps: hasTimestamps(fields),
+        };
```
VincentMolinie: Much better like that good job. We could in a next PR detect `deletedAt` to set the paranoid options too ^^

lumber-generate.js
```diff
@@ -77,6 +77,7 @@ program
       fields: schema[table].fields,
```
VincentMolinie: You could just pass `schema[table]` no ?

services/database-analyzer.js
```diff
@@ -17,6 +17,33 @@ function DatabaseAnalyzer(databaseConnection, config) {
     return Object.keys(schema).filter(column => schema[column].primaryKey);
   }
 
+  function isUnderscored(fields) {
+    let underscored = false;
```
VincentMolinie: we could simplify this like that:
`return fields.some(field => field.name.includes('_'));`

services/dumper.js
```diff
@@ -141,16 +114,16 @@ function Dumper(config) {
...
     const templatePath = `${__dirname}/../templates/model.txt`;
     const template = _.template(fs.readFileSync(templatePath, 'utf-8'));
 
     const text = template({
       table,
       fields,
       references,
-      underscored: isUnderscored(fields),
-      timestamps: hasTimestamps(fields),
+      underscored: options.underscored,
```
VincentMolinie: `...options` would be even better. So if we add a new options we don't have to add it here every time ;)

services/migrator.js
```diff
@@ -43,8 +16,8 @@ function Migrator(config) {
       table,
       fields,
       references,
-      underscored: isUnderscored(fields),
-      timestamps: hasTimestamps(fields),
+      underscored: options.underscored,
```
VincentMolinie: Same here `...options`

PR #247, opened 2019-08-13
[*] Command Generate - Fix generation of relationship if foreign key is unconventionnal

VincentMolinie:


VincentMolinie:
[Lumber - Tables with camelcased column names having at least one column with an '_' does not display if they have relationships.](https://trello.com/c/7r884X93/3605-lumber-tables-with-camelcased-column-names-having-at-least-one-column-with-an-does-not-display-if-they-have-relationships)

PR #255, opened 2019-09-09
[*]  Command Generate - Avoid foreignKey to conflict with relationship

VincentMolinie:
When foreignKey are generated, if they have the same name as the alias it crash the server on start

### DIFF
--- a/services/database-analyzer.js
+++ b/services/database-analyzer.js
@@ -79,7 +79,7 @@ function DatabaseAnalyzer(databaseConnection, config, allowWarning) {
             const reference = {
               ref: foreignKey.foreign_table_name,
               foreignKey: foreignKey.column_name,
-              foreignKeyname: _.camelCase(foreignKey.column_name),
+              foreignKeyName: _.camelCase(foreignKey.column_name),
               as: formatAliasName(foreignKey.column_name),
             };
 
--- a/services/database-analyzer.js
+++ b/services/database-analyzer.js
@@ -83,6 +83,8 @@ function DatabaseAnalyzer(databaseConnection, config, allowWarning) {
               as: formatAliasName(foreignKey.column_name),
             };
 
+            // NOTICE: If the foreign key name and alias are the same, Sequelize will crash, we need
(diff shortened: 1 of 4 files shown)


services/database-analyzer.js
```diff
@@ -79,9 +79,14 @@ function DatabaseAnalyzer(databaseConnection, config) {
             const reference = {
               ref: foreignKey.foreign_table_name,
               foreignKey: foreignKey.column_name,
+              foreignKeyname: _.camelCase(foreignKey.column_name),
```
arnaudbesnier:
```suggestion
              foreignKeyName: _.camelCase(foreignKey.column_name),
```
I guess

services/database-analyzer.js
```diff
@@ -79,9 +79,14 @@ function DatabaseAnalyzer(databaseConnection, config) {
...
               as: formatAliasName(foreignKey.column_name),
             };
 
+            if (reference.foreignKeyName === reference.as) {
+              reference.foreignKeyName = `${reference.foreignKeyName}Key`;
```
arnaudbesnier: I did a quick test about the `Key` suffix and it reinforces the fact that I am not confident with this change.
99% sure it will break some generated queries in the case of "conventional" column name.
Screenshot 2019-09-25 at 17 36 07

services/dumper.js
```diff
@@ -132,7 +132,7 @@ function Dumper(config) {
       const expectedConventionalForeignKeyName = underscored
         ? _.snakeCase(reference.foreignKey) : reference.foreignKey;
       const foreignKeyColumnUnconventional =
-        reference.foreignKey !== expectedConventionalForeignKeyName;
+        reference.foreignKeyName !== expectedConventionalForeignKeyName;
```
VincentMolinie: We mark a field as unconventional if it's not the same as foreignKeyName (To avoid the previous bug you had)

arnaudbesnier: We should have tests created to protect this code against regressions according to our new rules.

PR #258, opened 2019-09-17
[+] Generate command - Change default application port to prevent conflicts

VincentMolinie:


VincentMolinie:
[Lumber - Change default application port to prevent conflicts.](https://trello.com/c/EyYTbMjU/3701-lumber-change-default-application-port-to-prevent-conflicts)

PR #259, opened 2019-09-18
[-] Generate command - Add a silent mode to remove all logs except warning and error

VincentMolinie:


VincentMolinie:
[Lumber - Remove logs](https://trello.com/c/KgiO3qB5/3700-lumber-remove-logs)

services/logger.js
```diff
@@ -1,16 +1,32 @@
...
 
-function logLine(color, message) {
-  console.log(`${chalk[color]('>')} ${message}`);
-}
+class Logger {
+  constructor(silent) {
+    this.silent = silent;
+  }
+
+  _logMessage(message, force) {
```
larcin: [Linter] Do you really need to bypass the `"no-underscore-dangle"` linter rule ? Can't you just use `logMessage` or anything else which wouldn't require a linter exception ?

lumber-generate.js
```diff
@@ -10,10 +10,15 @@ const logger = require('./services/logger');
 program
   .description('Generate a backend application with an ORM/ODM configured.')
   .option('-c, --connection-url <connectionUrl>', 'Enter the database credentials with a connection URL')
+  .option('--silent', 'Do not print any logs')
```
arnaudbesnier: A new option?
Why don't we keep things simple and remove those logs.

+ it won't be silent as the "Hooray" log remains printed.

arnaudbesnier: We want to simplify Lumber in the near future, but such implementation is going to the opposite direction :/

PR #266, opened 2019-10-04
[*] Records Update - Fix `TINYINT` column update when declared as a `BOOLEAN` field in the ORM model

rap2hpoutre:
When trying to update a field in ForestAdmin that is a `TINYINT` in a MySQL DB (which is the actual MySQL type for boolean) and declared as a `boolean` in a model of the Lumber instance, we get the following error:

> Incorrect integer value: 'false' for column 'feature_1' at row 1

See: https://github.com/sequelize/sequelize/issues/10633

This is [now fixed](https://github.com/sequelize/sequelize/issues/10633#issuecomment-483786351) in [`mysql2` package](https://www.npmjs.com/package/mysql2). Since the version upgrade is minor, we can update the latest version of `mysql2` (`1.7.0`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 ### Fixed
 - Command Generate - Avoid foreignKey to conflict with relationship.
 - Command Generate - Fix creation of project containing whitespaces.
-- Technical - Fix MySQL boolean values (Tinyint)
+- Records Update - Fix `TINYINT` column update when declared as a `BOOLEAN` field in the ORM model.
 
 ## RELEASE 2.4.0 - 2019-09-20
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### Fixed
 - Command Generate - Avoid foreignKey to conflict with relationship.
 - Command Generate - Fix creation of project containing whitespaces.
+- Technical - Fix MySQL boolean values (Tinyint)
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "mkdirp": "^0.5.1",
     "mongodb": "3.3.0",
     "mysql": "^2.11.1",
-    "mysql2": "1.4.2",
+    "mysql2": "^1.7.0",
```
arnaudbesnier: We tend to set strict dependency version to reduce the uncertainty of the install dependency versions.
1. Using `^` should not be a problem if the dependencies follow the major/minor/patch rules but we cannot be sure of that.
2. I am not a fan of the idea that, depending of the date you install a package, it dependencies could change. (especially for debug purpose it could be painful)

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 ### Fixed
 - Command Generate - Avoid foreignKey to conflict with relationship.
 - Command Generate - Fix creation of project containing whitespaces.
+- Technical - Fix MySQL boolean values (Tinyint)
```
```suggestion
- Records Update - Fix `Tinyint` column update when declared as a `boolean` field in the ORM model.
```
Seems better.
This is not a "Technical" change as it may impact the user experience.

rap2hpoutre: Thanks. I also changed the case of data types since it seems to be uppercase in MySQL doc.

PR #271, opened 2019-10-08
[+] Command Generate - Add "Application host" parameter (`-H, --application-host`) to `lumber generate`

rap2hpoutre:
~~Waiting for #269 to be reviewed.~~

PR #273, opened 2019-10-10
[+] Command Generate - Add  "SSL" parameter (`-S, --ssl`) to `lumber generate`

rap2hpoutre:
See: https://github.com/ForestAdmin/lumber/pull/269 & https://github.com/ForestAdmin/lumber/pull/271. According to @arnaudbesnier suggestion, to avoid breaking changes, `--ssl` option is not a boolean, but a two value option ("true" or "false").

_I added a small fix. Previously, if the `DATABASE_SSL` had a wrong value, it crashed due to JSON parsing error. Now it just fails with a user friendly(?) message._


rap2hpoutre:
[Lumber - add --ssl param](https://trello.com/c/4iUKtDiB/3742-lumber-add-ssl-param)

PR #276, opened 2019-10-11
[+] Command generate - Add parameters documentation to `lumber install`

rap2hpoutre:
See: https://github.com/ForestAdmin/lumber-forestadmin/pull/10 (params are handled by _ForestAdmin/lumber-forestadmin_)

### DIFF
--- a/lumber-install.js
+++ b/lumber-install.js
@@ -12,8 +12,8 @@ const argv = require('minimist')(process.argv.slice(2));
 program
   .description('Install a Lumber plugin')
   .option('-e, --email <email>', 'Your Forest Admin account email')
-  .option('-P, --password <password>', 'Your Forest Admin account password')
-  .option('-t, --token <token>', 'Your Forest Admin account token')
+  .option('-P, --password <password>', 'Your Forest Admin account password (ignored if token is set)')
+  .option('-t, --token <token>', 'Your Forest Admin account token (required if token is missing)')
   .option('-p, --projectName <projectName>', 'Your Forest Admin project name')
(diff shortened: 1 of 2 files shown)


rap2hpoutre:
[Lumber - Add --token param](https://trello.com/c/cWcCD2EG/3752-lumber-add-token-param)

lumber-install.js
```diff
@@ -11,6 +11,10 @@ const argv = require('minimist')(process.argv.slice(2));
 
 program
   .description('Install a Lumber plugin')
+  .option('-e, --email <email>', 'Your Forest Admin account email')
+  .option('-P, --password <password>', 'Your Forest Admin account password')
+  .option('-t, --token <token>', 'Your Forest Admin account token')
```
Sidnioulz: Is that an authentication token for your forst admin password?

Are -P and -t mutually exclusive? If so, it might be useful to state it.

rap2hpoutre: @Sidnioulz 

> Are -P and -t mutually exclusive? If so, it might be useful to state it.

Yes, good catch! Not sure how to give this info though. Do you have any suggestion?

Sidnioulz: Something like that:

* -f, --foo <foo>, 'Does foo. Ignored if bar is set.'
* -b --bar <bar>, 'Does bar. When both foo and bar are set, foo is ignored.'

rap2hpoutre: Thank you! Fixed! (I adapted your text to reflect the actual behavior (which is a bit strange BTW))

PR #277, opened 2019-10-11
[*] Command Generate - Fix `mysql2` error with ssl as a boolean

rap2hpoutre:
🙄 `mysql2` has done some breaking change in minor revision (...semver). See: https://github.com/sidorares/node-mysql2/pull/896, https://github.com/request/request/issues/3073 and https://github.com/sidorares/node-mysql2/pull/895. I don't know why I did not see this issue when testing https://github.com/ForestAdmin/lumber/pull/268, I may have missed something 😕; I have to be more cautious.

Without this fix:

```
➜  lumber git:(devel) ./lumber.js generate baz -c "mysql://foo:bar@localhost:5435"
...
Ignoring invalid configuration option passed to Connection: encrypt. This is currently a warning, but in future versions of MySQL2, an error will be thrown if you pass an invalid configuration option to a Connection
> Cannot connect to the database due to the following error:
> SequelizeConnectionError: SSL profile must be an object, instead it's a boolean
```

With the fix:

```
➜  lumber git:(fix/mysql2-breaking-change) ./lumber.js generate baz -c "mysql://foo:bar@localhost:5435"
...
> Hooray, installation success!
```

_Side note: SSL option with MySQL has never worked according to the issues I read._


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
+- Command Generate - Fix `mysql2` error with ssl as a boolean.
```
arnaudbesnier: [Conventions] This is a technical description of the issue IMO. We should explain to the end users what usage regression has been fixed. Something like:
```suggestion
- Command Generate - Fix projects generation based on MySQL connections. (Regression introduced in v?)
```

templates/app/models/index.js
```diff
@@ -19,7 +19,11 @@ let databaseOptions = {
 };
 
 if (process.env.DATABASE_SSL && JSON.parse(process.env.DATABASE_SSL.toLowerCase())) {
-  databaseOptions.dialectOptions.ssl = true;
+  if (process.env.DATABASE_URL.startsWith('mysql://')) {
+    databaseOptions.dialectOptions.ssl = { rejectUnauthorized: true };
+  } else {
+    databaseOptions.dialectOptions.ssl = true;
+  }
```
arnaudbesnier: I am wondering if this is a good idea or not to insert conditions like `if mysql` in a generated server that will never use `mysql` databases.
Something like:
```javascript
<% if (config.dbDialect === 'mysql') {
%>databaseOptions.dialectOptions.ssl = { rejectUnauthorized: true };
<% } else {
%>databaseOptions.dialectOptions.ssl = true;
<% } %>
```

rap2hpoutre:
@arnaudbesnier Updated! The indentation is a bit weird in template, but that's the only way I found to generate a file with correct indentation (avoid unwanted line break)

CHANGELOG.md
rap2hpoutre: Thanks!

PR #279, opened 2019-10-14
[*] Command Generate - Remove schema prompt for MySQL

rap2hpoutre:
`lumber generate` ask for schema for MySQL install. It could be removed since it's useless, MySQL does not use schema. Source: https://stackoverflow.com/questions/11618277/difference-between-schema-database-in-mysql & https://dev.mysql.com/doc/refman/8.0/en/glossary.html#glos_schema

_Issue detected during new features tests_

### DIFF
--- a/services/prompter.js
+++ b/services/prompter.js
@@ -106,8 +106,7 @@ async function Prompter(program, requests) {
         when: (answers) => {
           // NOTICE: MongoDB, MySQL and SQLite do not require a Schema.
           const skipDatabases = ['sqlite', 'mongodb', 'mysql'];
-          return !(skipDatabases.includes(answers.dbDialect)
-            || skipDatabases.includes(envConfig.dbDialect));
+          return !skipDatabases.includes(answers.dbDialect || envConfig.dbDialect);
         },
         default: (args) => {
           if (args.dbDialect === 'postgres') { return 'public'; }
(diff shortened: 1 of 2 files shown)


rap2hpoutre:
[Lumber - Ensure all prompts are avoidable with all options set](https://trello.com/c/FHUmTryI/3750-lumber-ensure-all-prompts-are-avoidable-with-all-options-set)

services/prompter.js
```diff
@@ -281,10 +286,10 @@ async function Prompter(program, requests) {
           if (password) {
             if (FORMAT_PASSWORD.test(password)) { return true; }
             return '🔓  Your password security is too weak 🔓\n' +
-            ' Please make sure it contains at least:\n' +
-            '    > 8 characters\n' +
-            '    > Upper and lower case letters\n' +
-            '    > Numbers';
+              ' Please make sure it contains at least:\n' +
```
Sidnioulz: [Conventions] Whilst I'm personally strongly in favour of updating code style in every file touched by a PR, I believe we were told to do this in separate cards during our onboarding. It'd be good to validate the convention to follow with @SteveBunlon, @larcin or @VincentMolinie.

services/prompter.js
```diff
@@ -103,7 +103,12 @@ async function Prompter(program, requests) {
         name: 'dbSchema',
         message: 'What\'s the database schema? [optional]',
         description: 'Leave blank by default',
-        when: answers => answers.dbDialect !== 'sqlite' && answers.dbDialect !== 'mongodb' && envConfig.dbDialect !== 'mongodb',
+        when: (answers) => {
+          // NOTICE: MongoDB, MySQL and SQLite do not require a Schema.
+          const skipDatabases = ['sqlite', 'mongodb', 'mysql'];
+          return !(skipDatabases.includes(answers.dbDialect)
```
Sidnioulz: Maybe use `answer.dbDialect || envConfig.dbDialect`. If someone has a lingering .bashrc or .env file setting a MongoDB / MySQL db but now tries to generate a PGSQL project, they wouldn't be prompted for the schema with the current code. If the answer is set at runtime, the env variable should be ignored.

rap2hpoutre: Good catch!

PR #280, opened 2019-10-14
[+] Commands - Add deprecation warning for environment variables usage

rap2hpoutre:
Displays something like this:

```
➜ DATABASE_URL=foo ./lumber.js generate bar
> Environment variables usage detected:
>  - DATABASE_URL
> Environment variables are DEPRECATED and will be removed in the future major Lumber version. Please use command parameters instead.
> Type `lumber help` for more information.
```

### DIFF
--- a/lumber-generate.js
+++ b/lumber-generate.js
@@ -6,6 +6,7 @@ const DatabaseAnalyzer = require('./services/database-analyzer');
 const Dumper = require('./services/dumper');
 const CommandGenerateConfigGetter = require('./services/command-generate-config-getter');
 const logger = require('./services/logger');
+const EnvironmentChecker = require('./services/environment-checker');
 
 program
   .description('Generate a backend application with an ORM/ODM configured')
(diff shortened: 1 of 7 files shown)


rap2hpoutre:
[Lumber - Deprecation warning](https://trello.com/c/yezLp233/3753-lumber-deprecation-warning)

lumber-generate.js
```diff
@@ -19,6 +20,23 @@ program
   .parse(process.argv);
 
 (async () => {
+  // NOTICE: Check deprecated environments variables.
```
Sidnioulz: In line with what Arnaud told me in past reviews, Forest prefers to avoid comments that feel redundant with the code. Here, you could name the `EnvironmentChecker` `EnvironmentDeprecationChecker` and it would be obvious why this is used for :-)

services/environment-checker.js
```diff
@@ -0,0 +1,19 @@
+function EnvironmentChecker(env, logger, deprecated) {
```
Sidnioulz: The only use you make of this service is to call `logWarnings` to check for inclusion in a list of active environment variables.

Because of this, I feel this could be a singleton service instead of an instantiable one, which directly exports the `logWarnings` method (e.g. how forestadmin/utils/*.js files work). Then, you could have something as direct as `warnOfDeprecatedEnvironment([...]);` in your scripts.

services/environment-checker.js
```diff
@@ -0,0 +1,19 @@
...
+  this.deprecated = deprecated;
+
+  this.getActives = () => this.deprecated.filter(variable => env[variable] !== undefined);
+
+  this.logWarnings = () => {
+    const activeEnvironmentVariables = this.getActives();
+    if (activeEnvironmentVariables.length) {
+      logger.warn(
+        'Environment variables usage detected:',
```
Sidnioulz:
```suggestion
        'Environment variable usage detected:',
```

services/environment-checker.js
```diff
@@ -0,0 +1,19 @@
...
+        activeEnvironmentVariables.map(variable => ` - ${variable}`),
```
Sidnioulz: Is it on purpose that you don't insert new lines?

rap2hpoutre: Hum, no 😅. Fixed. Thank you!

PR #283, opened 2019-10-14
[-] Command Generate - Give models a unique pascal-cased name when generating sequelize model files

Sidnioulz:
…nerating sequelize model files


Sidnioulz:
[Lumber - Replace "Model" in models templates by the real model name to be more conventional / dev friendly](https://trello.com/c/IrFUu2B2/3762-lumber-replace-model-in-models-templates-by-the-real-model-name-to-be-more-conventional-dev-friendly)

arnaudbesnier:
> Approved! 🚀
> 
> Could we have avoided: `<%= _.chain(table).camelCase().upperFirst().value() %>` repetition?

The question itself is enough not to validate this PR.

Sidnioulz:
Can't argue with that. Sent a fix.

templates/model.txt
```diff
@@ -9,9 +9,13 @@ const schema = mongoose.Schema({<% _.each(fields, (field) => { %>
 
 module.exports = mongoose.model('<%= table %>', schema, '<%= table %>');
 <% } else {
+%><%
+const pascalCase = function(input) {
+  return _.chain(input).camelCase().upperFirst().value();
```
arnaudbesnier: [Code style] This logic should not be defined in the template, but in the services that call the template (`dumper.js` & `migrator.js`) to make the template silly.
It will also make the template easier to read.

PR #287, opened 2019-10-17
[*] Continuous Integration - Change `11.14` version of `node_js` to ` lts/*` in `.travis.yml`

rap2hpoutre:
11.14 is abandonned since July 2019: https://nodejs.org/en/about/releases/. We have to ensure CI use the current LTS version to build/test Lumber.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 
 language: node_js
 
-node_js: stable
+node_js: lts/*
 
 env:
   - NODE_ENV=travis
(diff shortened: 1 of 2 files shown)


rap2hpoutre:
[Lumber - Run CI on an up to date node version](https://trello.com/c/Jp53VSrK/3798-lumber-run-ci-on-an-up-to-date-node-version)

rap2hpoutre:
> As this been discussed with anybody before ? 

No, only with you.

> Do we really want this to be tested with the latest and not a fixed version on which we could guarantee at 100% that lumber will work properly ?

Yes. Latest LTS (with security fix) is better IMO than abandoned 11.4. We could add more versions later (including current)

PR #291, opened 2019-10-21
[*] Technical - Fix repository in `package.json`

rap2hpoutre:
To avoid this warning:

```
➜  lumber git:(fix/lodash-version) ✗ npm i lodash@4.17.15 --save-exact
...
npm WARN lumber-cli@2.5.1 No repository field.
...
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "Valentin Lamatte <valentinl@forestadmin.com>",
     "Arnaud Besnier <arnaudibesnier@gmail.com>"
   ],
+  "repository": "https://github.com/ForestAdmin/lumber",
   "license": "MIT",
   "dependencies": {
     "bluebird": "^3.4.6",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "Valentin Lamatte <valentinl@forestadmin.com>",
     "Arnaud Besnier <arnaudibesnier@gmail.com>"
   ],
+  "repository": "https://github.com/ForestAdmin/lumber",
```
SteveBunlon: [ Improvement ] As stated in the doc, repository has to be used with 'type' ("type" : "git"). If you want to use the shortcut (as you did) it's better to write that way : "repository": "github:ForestAdmin/lumber".

See ref : https://docs.npmjs.com/files/package.json#repository

rap2hpoutre: Thank you! 👍 I should have read the doc 😅

PR #293, opened 2019-10-22
[*] Command Generate - Fix foreign key imports using connection URL with MySQL

rap2hpoutre:
I do not usually like PR that fixes two issues, but in this case, fixing one has revealed one other: I would have not seen the foreign key error without Sequelize replacements. So the two fix are relatives to each other. So basically:
 - `field = '${foo}'` as been replaced by `field = :foo`+`{ replacements: { foo } }`
 - In case there is no `dbName`, `connectionUrl` is parsed for MySQL (which is the only one who requires its own database name when analyzing foreign keys).

### DIFF
--- a/services/table-foreign-keys-analyzer.js
+++ b/services/table-foreign-keys-analyzer.js
@@ -3,6 +3,7 @@ function TableForeignKeysAnalyzer(databaseConnection, config) {
 
   this.perform = async (table) => {
     let query = null;
+    let replacements = { table };
 
     switch (queryInterface.sequelize.options.dialect) {
       case 'postgres':
@@ -19,7 +20,7 @@ function TableForeignKeysAnalyzer(databaseConnection, config) {
           JOIN information_schema.constraint_column_usage AS ccu
             ON ccu.constraint_name = tc.constraint_name
           WHERE constraint_type = 'FOREIGN KEY'
-            AND tc.table_name='${table}';`;
+            AND tc.table_name=:table;`;
         break;
       case 'mysql':
         query = `
@@ -29,8 +30,11 @@ function TableForeignKeysAnalyzer(databaseConnection, config) {
             REFERENCED_TABLE_NAME AS foreign_table_name,
             REFERENCED_COLUMN_NAME AS foreign_column_name
           FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE
-          WHERE TABLE_SCHEMA = '${config.dbName}'
-            AND TABLE_NAME = '${table}';`;
+          WHERE TABLE_SCHEMA = :databaseName
+            AND TABLE_NAME = :table;`;
+        // NOTICE: `config.dbName` may not exists when using connection URL. 
+        //         Use `dbConnectionUrl` as a fallback.
+        replacements.databaseName = config.dbName || config.dbConnectionUrl.split("/").pop();
         break;
       case 'mssql':
         query = `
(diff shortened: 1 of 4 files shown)


services/table-foreign-keys-analyzer.js
```diff
@@ -29,8 +30,11 @@ function TableForeignKeysAnalyzer(databaseConnection, config) {
             REFERENCED_TABLE_NAME AS foreign_table_name,
             REFERENCED_COLUMN_NAME AS foreign_column_name
           FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE
-          WHERE TABLE_SCHEMA = '${config.dbName}'
-            AND TABLE_NAME = '${table}';`;
+          WHERE TABLE_SCHEMA = :databaseName
+            AND TABLE_NAME = :table;`;
+        // NOTICE: `config.dbName` may not exists when using connection URL.
+        //         Use `dbConnectionUrl` as a fallback.
+        replacements.databaseName = config.dbName || config.dbConnectionUrl.split('/').pop();
```
VincentMolinie: [Spec] This doesn't work with a custom schema `postgres://user:password@localhost:5432/my-db?schema=oops` And the schema can also contains slash !

VincentMolinie: WHy not just retrieving the database name from sequelize 🤔  ?

rap2hpoutre: Good catch! Thanks!

services/table-foreign-keys-analyzer.js
```diff
@@ -49,17 +51,18 @@ function TableForeignKeysAnalyzer(databaseConnection, config) {
...
         break;
       case 'sqlite':
-        query = `PRAGMA foreign_key_list('${table}');`;
+        query = 'PRAGMA foreign_key_list(:table);';
         break;
       default:
         break;
     }
 
+    console.log(replacements);
```
VincentMolinie: [Code Cleaning] Self review 😛 ?

services/table-foreign-keys-analyzer.js
```diff
@@ -29,8 +30,9 @@ function TableForeignKeysAnalyzer(databaseConnection, config) {
             REFERENCED_TABLE_NAME AS foreign_table_name,
             REFERENCED_COLUMN_NAME AS foreign_column_name
           FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE
-          WHERE TABLE_SCHEMA = '${config.dbName}'
-            AND TABLE_NAME = '${table}';`;
+          WHERE TABLE_SCHEMA = :databaseName
+            AND TABLE_NAME = :table;`;
+        replacements.databaseName = queryInterface.sequelize.config.database;
```
VincentMolinie: [MISC] Much better 👍

VincentMolinie: But when I look at the doc I see `getDatabaseName()` which seems more friendly and public to me don't you think 🤔 ?
https://sequelize.org/master/class/lib/sequelize.js~Sequelize.html

services/table-foreign-keys-analyzer.js
rap2hpoutre: woops.

services/table-foreign-keys-analyzer.js
VincentMolinie: Ahah it's sequelize....

rap2hpoutre:
[Lumber - fix foreign key bug with connexion URL for MySQL](https://trello.com/c/Hjn3Ys2u/3807-lumber-fix-foreign-key-bug-with-connexion-url-for-mysql)

PR #298, opened 2019-10-24
[-] Database analyzer refacto - each analyzer in its file

slimee:


slimee:
This PR consist of function moving to separate concerns - no feature has changed.
a small Error management has been added too.

services/database-analyzer-sequelize.js
```diff
@@ -0,0 +1,171 @@
...
+    if (typeof table === 'object') {
+    // eslint-disable-next-line no-param-reassign
+      table = table.tableName;
+    }
+
+    schema[table] = await analyzeTable(table, config);
+  });
+
+  if (_.isEmpty(schema)) {
+    throw new DatabaseAnalyzerError.EmptyDatabase('sequelize', databaseConnection.getDialect());
```
arnaudbesnier: [Regression?] I don't know if you tested it or not but it looks broken. The catch won't understand the `details.orm` and `details.dialect` from what I can see.

And to confirm that, the mongoose case seems to have the right arguments according to your new implementation.
```javascript
new DatabaseAnalyzerError.EmptyDatabase('no collections found', {
  orm: 'mongoose',
  dialect: 'mongodb',
});
```

services/database-analyzer-sequelize.js
```diff
@@ -0,0 +1,171 @@
+const ColumnTypeGetter = require('./column-type-getter');
+const TableForeignKeysAnalyzer = require('./table-foreign-keys-analyzer');
+const _ = require('lodash');
+const logger = require('./logger');
+const P = require('bluebird');
```
arnaudbesnier: [Conventions?] For readability I would move dependency require on top and internal code after that.
``` javascript
const P = require('bluebird');
const _ = require('lodash');
const logger = require('./logger');
const ColumnTypeGetter = require('./column-type-getter');
const TableForeignKeysAnalyzer = require('./table-foreign-keys-analyzer');
const { DatabaseAnalyzerError } = require('../utils/errors');
```

utils/errors.js
```diff
@@ -0,0 +1,20 @@
+const VError = require('verror');
```
arnaudbesnier: I have never really understood the point of `verror` dependency. I am not sure this is interesting here.
But if you think so, ok.

services/database-analyzer-mongo.js
```diff
@@ -0,0 +1,104 @@
...
+
+function isUnderscored(fields) {
+  return fields.every(field => field.nameColumn === _.snakeCase(field.nameColumn))
+    && fields.some(field => field.nameColumn.includes('_'));
+}
+
+function analyzeMongoCollection(databaseConnection, collectionName) {
+  return new Promise((resolve, reject) => {
+    databaseConnection.collection(collectionName)
+    /* eslint-disable */
```
arnaudbesnier: [Code Style] Do you still need this comment? If so, you should reindent it to make the code easy to read (and like it was before)

package.json
```diff
@@ -37,6 +37,7 @@
     "sequelize": "git://github.com/ForestAdmin/sequelize.git#feature/support-schema",
     "superagent": "5.1.0",
     "tedious": "6.4.0",
+    "verror": "^1.10.0",
```
arnaudbesnier: Strict versions are really more "secure".
Using `^` can make the dependencies version different across time which is bad for me.
A developer could have a different experience than another because of this.

Also we cannot be sure that `verror` dependencies are properly applying semantic versioning.
They could introduce breaking changes in "patch" version... (Exactly like I badly did for Lumber v2.5.1)

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 ### Changed
 - Docker - Fix lumber version in `Dockerfile`.
+- Technical - database analyzer refacto - separate each database analyzer in its own file and add errors.
```
```suggestion
- Technical - Split database analysis logic in 2 separated files (SQL and NoSQL).
```

services/database-analyzer.js
```diff
@@ -1,265 +1,38 @@
...
+  return process.exit(1);
+}
 
-    if (_.isEmpty(schema)) {
-      return reportEmptyDatabase('sequelize', databaseConnection.getDialect());
+function DatabaseAnalyzer(databaseConnection, config, allowWarning) {
+  this.perform = async () => {
+    let analyze;
+    if (config.dbDialect === 'mongodb') {
+      analyze = mongoTableAnalyzer;
```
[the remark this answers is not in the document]
slimee: `mongoTableAnalyzer` & `sequelizeTableAnalyzer` are function, yes. 
To avoid the weird look assignation, we may spell them `mongoAnalyze` & `sequelizeAnalyze`?

I understand the point about the inconsistent naming of the two analyzers, one is orm and one is DB. It's not the same level.
But I don't feel agree because they are concrete implementations. They should be named by what they do, at the lower level. I figure that because if we add another NoSql DB tomorrow (let's say Cassandra, CouchBase, Hazelcast or any), we won't be able to name the analyzer. We may also add other ways to analyze SQL, too?

utils/errors.js
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  util.inherits(LumberError, VError);
+
+  return LumberError;
+}
+
+exports.DatabaseAnalyzerError = {
+  EmptyDatabase: createLumberError('EmptyDatabase'),
+};
```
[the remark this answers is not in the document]
slimee: What a sniper eye :)

services/database-analyzer-sequelize.js
slimee: Oh yes, the parameters are not OK. 
I tested it, and it displays:
`> Your database looks empty! Please create some tables before running the command.`

That was OK for me but you are right, it contains a regression.
Thanks

services/database-analyzer-sequelize.js
slimee: Totally agree. Thanks & sorry

utils/errors.js
slimee: I think it's useful when you set a hierarchy of errors.
We may remove this from now, I am ok with that

CHANGELOG.md
slimee: Let me know your advice about the analyzer's names above before changing here.

services/database-analyzer.js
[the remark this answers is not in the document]
arnaudbesnier: all right, then, go for:
`mongoCollectionsAnalyze` & `sequelizeTablesAnalyze`
or
`analyzeMongoCollections` & `analyzeSequelizeTables`

And lets reflect this choice in the file names and `exports`.

👍

CHANGELOG.md
arnaudbesnier: I answered for the analyzers names, up to you know ;)

services/database-analyzer.js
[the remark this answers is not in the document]
slimee: 👍

CHANGELOG.md
slimee: okay

PR #309, opened 2019-10-29
[+] Field Types - Support `TIMESTAMP(n)`, `SMALLINT(n)` and `BIGINT(n)` column types (MySQL)

rap2hpoutre:
Some MySQL types are ignored on generation. Please test this carefully (it should not break existing types recognition).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 ### Added
-- Field Types - Support `TIMESTAMP(n)`, `SMALLINT(n)`, `BIGINT(n)` and `BIT(1)` column types (MySQL)
+- Field Types - Support `TIMESTAMP(n)`, `SMALLINT(n)` and `BIGINT(n)` column types (MySQL)
 
 ## RELEASE 2.6.3 - 2019-11-04
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Field Types - Support `TIMESTAMP(n)`, `SMALLINT(n)`, `BIGINT(n)` and `BIT(1)` column types (MySQL)
 
 ## RELEASE 2.6.3 - 2019-11-04
 ### Fixed
--- a/services/column-type-getter.js
+++ b/services/column-type-getter.js
@@ -57,7 +57,6 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
 
     switch (type) {
       case 'BIT': // NOTICE: MSSQL type.
-      case 'BIT(1)': // NOTICE: MySQL type.
       case 'BOOLEAN':
         return 'BOOLEAN';
       case 'CHARACTER VARYING':
--- a/services/column-type-getter.js
+++ b/services/column-type-getter.js
@@ -105,8 +105,6 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
       case 'DATE':
       case 'DATETIME':
       case (type.match(/^TIMESTAMP.*/i) || {}).input:
-      case 'TIMESTAMP WITH TIME ZONE':
-      case 'TIMESTAMP WITHOUT TIME ZONE':
         return 'DATE';
       case 'TIME':
       case 'TIME WITHOUT TIME ZONE':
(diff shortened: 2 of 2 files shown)


services/column-type-getter.js
```diff
@@ -103,7 +104,7 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
         return 'DOUBLE';
       case 'DATE':
       case 'DATETIME':
-      case 'TIMESTAMP':
+      case (type.match(/^TIMESTAMP.*/i) || {}).input:
       case 'TIMESTAMP WITH TIME ZONE':
       case 'TIMESTAMP WITHOUT TIME ZONE':
```
rap2hpoutre: We never reach the last two cases because of the previous line. I kept them to be explicit, but maybe we have to remove them. Let me know what's better!

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Field Types - Support `TIMESTAMP(n)`, `SMALLINT(n)`, `BIGINT(n)` and `BIT(1)` column types (MySQL)
```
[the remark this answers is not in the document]
rap2hpoutre: @arnaudbesnier Tested locally: I installed with lumber generate (using this branch) then added+changed entries in my database from Forest Admin interface (ugly screenshot below). **Still, I actually think the reviewer should test it too!** I'm not 100% confident.

Capture d’écran 2019-10-29 à 17 08 08

services/column-type-getter.js
VincentMolinie: Yes Remove them

rap2hpoutre: @VincentMolinie Ok done!

services/column-type-getter.js
```diff
@@ -57,6 +57,7 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
 
     switch (type) {
       case 'BIT': // NOTICE: MSSQL type.
+      case 'BIT(1)': // NOTICE: MySQL type.
```
VincentMolinie: [Spec] This was removed for a good reason, this doesn't works on postgresql 
Capture d’écran 2019-10-31 à 14 10 19
I did not test on mysql but the argument is already enough to remove this line or find another way. By the way if I remember on some version of mysql (Because I worked on it 2 months ago I think) it was not working

rap2hpoutre: OK, thank you for testing this! 👍

> this doesn't works on postgresql

Good catch. Just for my understanding, what's the purpose of using `bit(1)` in postgreSQL since `boolean` exists? Anyway, I have to fix it!

> By the way if I remember on some version of mysql (Because I worked on it 2 months ago I think) it was not working

Could we find some history about this? Maybe it's fixed now: `sequelize` and `mysql2` were both upgraded.

Anyway, I have several options:
 - Revert my work about `bit(1)` (and open a new issue to fix this and test it with various MySQL version)
 - Consider it work with MySQL and modify the code to only consider this case for MySQL
 - something else.

What is your opinion @VincentMolinie ?

services/column-type-getter.js
```diff
@@ -82,14 +83,14 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
         return 'UUID';
       case 'JSONB':
         return 'JSONB';
-      case 'SMALLINT':
       case 'INTEGER':
       case 'SERIAL':
       case 'BIGSERIAL':
       case (type.match(/^INT.*/i) || {}).input:
+      case (type.match(/^SMALLINT.*/i) || {}).input:
       case (type.match(/^TINYINT.*/i) || {}).input:
```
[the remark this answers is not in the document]
rap2hpoutre: > kind of ugly

Rude 😳.

I used the same code snippet that was used everywhere, I did not want to add refactoring in this PR. I feel like this whole part could be refactored (I find it not elegant too), but I think smaller PR that answer only one subject are better (easier to understand, test, review and document). Maybe I'm wrong, I am not totally aware of our work process.

So I prefer ignore this comment for this actual PR. What is your opinion?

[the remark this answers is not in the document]
VincentMolinie: Ahah I was talking about the old code> I know this is not your fault 😉. Sorry ahah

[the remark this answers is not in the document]
VincentMolinie: Hmm leav it like that for now but we should definitely clean it ^^

services/column-type-getter.js
VincentMolinie: I would prefer reverting the bit(1) and doing another PR for this issue as it is much harder/longer than the rest of the PR

PR #295, opened 2019-10-23
[-] Commands Generate - Now also install forestadmin middleware

VincentMolinie:

PR #310, opened 2019-10-30
[*] Command Login - Fix login to match old behaviour

VincentMolinie:
As you mentioned here: https://github.com/ForestAdmin/lumber/pull/295, the login has regressed. The purpose of this PR is to implement the same behaviour as before and adding the new parameters

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### Fixed
 - Command Login - Disable account creation.
-- Command Login - Verify the validity of the token passed in paramaters.
+- Command Login - Check token validity before login.
 
 ## RELEASE 2.6.1 - 2019-10-30
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 
 ### Fixed
 - Technical - Fix repository in `package.json`.
+- Command Login - Fix login to match old behaviour.
 
 ## RELEASE 2.5.2 - 2019-10-25
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 - Commands - Remove command run and install.
 - Command Generate - Install Forest Admin middleware.
 
+### Fixed
+- Command Login - Fix login only enable connection (And not account creation).
+- Command Login - Verify the validity of the token passed in paramaters.
+
 ## RELEASE 2.6.1 - 2019-10-30
 ### Fixed
 - Command Generate - Fix install on mssql when there is two tables with the same name in different schemas.
@@ -24,7 +28,6 @@
 
 ### Fixed
 - Technical - Fix repository in `package.json`.
-- Command Login - Fix login to match old behaviour.
 
 ## RELEASE 2.5.2 - 2019-10-25
 ### Changed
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - Commands - Remove command run and install.
 - Command Generate - Install Forest Admin middleware.
 
+### Fixed
+- Command Login - Fix login only enable connection (And not account creation).
+- Command Login - Verify the validity of the token passed in paramaters.
```
VincentMolinie: If you have better ideas on the changelog, feel free

rap2hpoutre:
```suggestion
- Command Login - Check token validity before login.
```

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - Commands - Remove command run and install.
 - Command Generate - Install Forest Admin middleware.
 
+### Fixed
+- Command Login - Fix login only enable connection (And not account creation).
```
```suggestion
- Command Login - Disable account creation.
```

services/authenticator.js
```diff
@@ -12,9 +12,29 @@ const logger = require('./logger');
...
 function Authenticator() {
+  this.pathToLumberrc = `${os.homedir()}/.lumberrc`;
+
+  this.saveToken = token => fs.writeFileSync(this.pathToLumberrc, token);
+
+  this.isTokenCorrect = (email, token) => {
+    const sessionInfo = parseJwt(token);
+    if (sessionInfo) {
+      if ((sessionInfo.exp * 1000) <= Date.now()) {
+        logger.warn('You tried to use a token an expired token');
```
rap2hpoutre:
```suggestion
        logger.warn('Your token has expired.');
```

services/authenticator.js
```diff
@@ -12,9 +12,29 @@ const logger = require('./logger');
...
+        return false;
+      }
+
+      if (sessionInfo.data.data.attributes.email === email) {
+        return true;
+      }
+      logger.warn('You tried to use a token of another account');
```
rap2hpoutre:
```suggestion
      logger.warn('Your token is invalid.');
```

I think we should stay vague in this catch-all case. BTW, I'm not sure this message is always true: a user that as a broken session data may be in that case too.

> _User type "p@ssw0rd"..._
> You tried to use a password of another account
> _Wohoo, I stole something and the system helps me understand how to do the next step!_

utils/authenticator-helper.js
```diff
@@ -12,7 +13,7 @@ function parseJwt(token) {
 
     return JSON.parse(jsonPayload);
   } catch (error) {
-    console.error('Session token parsing unexpected error:', error);
+    logger.error('Your session token seems to be wrong');
```
rap2hpoutre:
```suggestion
    logger.error('Your session token is invalid.');
```

services/authenticator.js
VincentMolinie: Yeah but you need to have the token 🤔. You can't find it magically

VincentMolinie:
> dmin authentication token: [input is hidden]
> >> Invalid token. Please enter your authentica

- I added a validation for emails like the one on the server ;)
- Hmm and `CTRL + C` for me is a normal behaviour when trying to exit. Furthermore, there is a copy button so it should be pretty easy to copy

VincentMolinie:
FOr now I would leave the code like that. The incoming refacto of the prompter should easily refactorize these line of code 😉

PR #311, opened 2019-10-30
[*] Command Generate - Fix install on mssql when there is two tables with the same name in different schemas

VincentMolinie:


VincentMolinie:
[P1 - Lumber - Having two tables with the same name on MSSQL crash the generate](https://trello.com/c/irkNUwZv/3832-p1-lumber-having-two-tables-with-the-same-name-on-mssql-crash-the-generate)

PR #312, opened 2019-10-31
[+] Command Generate - Add tests on model generation (MySQL and PostgreSQL)

rap2hpoutre:
This PR adds some basic tests on MySQL and PostgreSQL [model generation](https://github.com/ForestAdmin/lumber/blob/devel/services/analyze-sequelize-tables.js). It's highly inspired by @slimee's [work](https://github.com/ForestAdmin/lumber/pull/296) (thank you ❤️).

It only test a few things: I prefer to keep it simple for the first round (for various reasons, the most important one is that I don't want to create something too complex if the review fails). In short, it tests two tables with relations for MySQL and PostgreSQL.

(Rooms for improvement) In order to entirely test the generate command, we could then:
 - Test the dumper (https://github.com/ForestAdmin/lumber/blob/devel/services/dumper.js)
 - Add more tests for more complex relations
 - Add more tests about column types.

I choose to separate MySQL and PostgreSQL fixtures and expected results intentionnaly. I feel like it's better to consider these two database as different (they _actually differs_). Sequelize is the shared lib, but database (and their lib) differs. Without that I have to create hooks and a bunch of `if`. 

Feel free to criticize and suggest improvements 😄

_To test it locally, don't forget to `docker-compose up`_.


rap2hpoutre:
[(8) Lumber Improvements - Add tests to avoid regression on generate command](https://trello.com/c/ex6J1EBk/3836-8-lumber-improvements-add-tests-to-avoid-regression-on-generate-command)

test/utils/sequelize-helper.js
```diff
@@ -0,0 +1,51 @@
...
+      },
+    });
+    return new Promise((resolve, reject) => {
+      this.sequelize.authenticate()
+        .then(() => resolve(this.sequelize))
+        .catch(err => reject(err));
+    });
+  }
+
+  async given(tableName) {
```
slimee: [IMO] I would use fixture term. `given a fixture` make more sense to me.

rap2hpoutre: In that particular case, it only accepts a table name (my `given` functions works actually only with a table name). That's far from perfect, but I felt like it's OK.

PR #315, opened 2019-11-05
Run tests on multiples environments (db versions)

slimee:


.env-test-max-version.ci
```diff
@@ -0,0 +1,3 @@
+DEVOPS_SLACK_CHANNEL=
+DEVOPS_SLACK_TOKEN=
```
VincentMolinie: [MISC] What is this doing here ?

.env-test-min-version.ci
```diff
@@ -0,0 +1,3 @@
+DEVOPS_SLACK_CHANNEL=
+DEVOPS_SLACK_TOKEN=
```
VincentMolinie: [MISC] What is this doing here ?

.env-test-max-version.ci
```diff
@@ -0,0 +1,3 @@
...
+MONGO_URL=mongodb://localhost:27016
```
VincentMolinie: [Code Convention] Missing line return at EOF

.travis.yml
```diff
@@ -16,11 +16,11 @@ services:
 before_install:
   - docker-compose --version
   - docker-compose up -d
-  - docker ps
+  - docker ps -a
```
VincentMolinie: [MISC] Why ?

package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "lint": "./node_modules/eslint/bin/eslint.js ./*.js ./assets ./deserializers ./serializers ./services ./test",
     "test": "./node_modules/mocha/bin/mocha 'test/**/*.js'",
```
VincentMolinie: [Specs] The yarn test command seems to be broken 🤔

package.json
```diff
@@ -6,6 +6,7 @@
...
+    "test:ci": "ENV=.env-test-min-version.ci npm run test && ENV=.env-test-max-version.ci npm run test",
```
VincentMolinie: [Specs] There is something that I really don't like in this solution is that not all the test needs to have different environment. Why not just having multiple environment variables and in the test do something like a foreach (Or an helper to do it) Like that we will have the context in the `describe` and  we don't have to run twice the test for the tests that does not need it

.travis.yml
slimee: Good to debug when a container is nok

package.json
slimee: What was the problem?

package.json
slimee: That' a good point indeed.

slimee: We will think about it before continue 👍

slimee: I've just produced a new version where tests are categorized and better selected 👍

package.json
```diff
@@ -6,6 +6,8 @@
   "scripts": {
     "lint": "./node_modules/eslint/bin/eslint.js ./*.js ./assets ./deserializers ./serializers ./services ./test",
     "test": "./node_modules/mocha/bin/mocha 'test/**/*.js'",
+    "test:int": "./node_modules/mocha/bin/mocha 'test/integration/**/*.js'",
+    "test:ci": "ENV=.env-test npm run test && ENV=.env-test-old-db npm run test:int",
```
VincentMolinie: I still really don't like that and it's harder to know on which database you are. And if you want to add one  more version you need to change this command line every time.
We could do something like that: https://github.com/ForestAdmin/lumber/pull/318
It does a simple describe and it's pretty easy to use/extends
Capture d’écran 2019-11-07 à 11 19 56

PR #316, opened 2019-11-05
[*] MSSQL - Fix connection issue using the ssl option.

SteveBunlon:


SteveBunlon:
[P1 - MS SQL docker onboarding - cannot connect to database due to encrypt config option issue](https://trello.com/c/3Xpf8VI7/3839-p1-ms-sql-docker-onboarding-cannot-connect-to-database-due-to-encrypt-config-option-issue)

PR #320, opened 2019-11-07
[*] Command Generate - Prompt for password/token just after the other prompts

VincentMolinie:


VincentMolinie:
[Lumber Improvement - Prompt for the password/token should be asked earlier](https://trello.com/c/GuMmMUGL/3846-lumber-improvement-prompt-for-the-password-token-should-be-asked-earlier)

PR #327, opened 2019-11-14
[+] Connection String - we encode credentials when possible.

slimee:
Please challenge this PR. It seems useless to me after doing it. (Am I tired?)

Given a connection string having `['@', '/', ':', '?']` in credentials part
When lumber generates
Then an error is thrown

Given a connection string with any special character except`['@', '/', ':', '?']` inside credentials part
When lumber generates
Then user and password must be url encoded.

But it seems customers have problem with `['@', '/', ':', '?']` characters only. These should be encoded.

Other special characters `=€-°'$&+,;()[]|~%*<>#^£^¨ù` can already be successfully used in user and password connection string. (mongo only)

1: People having ['@', '/', ':', '?']` in its credentials must encode them. (this PR says that)
2: What about people already encoding their values? Encoding twice will fail the auth.

**1+2 = people with `['@', '/', ':', '?']` in credentials can not auth at all.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- ConnectionString - credentials are uri encoded when possible.
+
 ### Changed
 - Technical - Remove unused ENCRYPT env variable.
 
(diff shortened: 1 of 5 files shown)


slimee:
[Onboarding - Mongo - encode passwords](https://trello.com/c/dpj9u4gX/3874-onboarding-mongo-encode-passwords)

slimee:
We should encode credentials typed from the UI

PR #328, opened 2019-11-15
[+] Command generate - mongo hasMany

slimee:


slimee:
[Lumber - Generate command creates hasMany relationships on NoSQL databases](https://trello.com/c/upSw1D9C/3871-lumber-generate-command-creates-hasmany-relationships-on-nosql-databases)

services/analyze-mongo-collections.js
```diff
@@ -14,11 +15,22 @@ const mapReduceOptions = {
   limit: 100,
 };
 
-// eslint-disable no-undef
+/* eslint-disable */
```
DrRaider: Why disable the linter here ?
Replace `var` by `let` or `const`, and use `Object.{keys, values, entries}` and iterate over the resulting array as suggested.

slimee: In this file, `mapCollection` and `reduceCollection` are not conventionnal functions. This two pieces of code run on the mongoDB side.
This is done this the use of the mongoDB `mapReduce` function.

This is the reason why we have `var` and other old school things: mongoDB is not compatible with last JS versions.

DrRaider: Maybe just disable the specific linter rules for this block of code, if we modify it later, we might miss other rules if we disable everything ;)

slimee: good idea!

PR #330, opened 2019-11-18
[+] Node version - Define supported node versions (10 and 12) in `package.json`

rap2hpoutre:
We support latests 10 and 12. 14 is not ready yet, 8 is dropped in one month.

Capture d’écran 2019-11-18 à 10 58 36

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,9 @@ sudo: required
 
 language: node_js
 
-node_js: lts/*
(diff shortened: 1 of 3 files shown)


rap2hpoutre:
[Fix node/npm versions on Lumber](https://trello.com/c/jizg7oqF/3877-fix-node-npm-versions-on-lumber)

PR #332, opened 2019-11-19
[*] Command Generate - Small fixes

VincentMolinie:


services/dumper.js
```diff
@@ -34,14 +34,15 @@ function Dumper(config) {
...
-      express: '~4.16.3',
+      chalk: '~1.1.3',
+      'cookie-parser': '1.4.4',
       debug: '~4.0.1',
       dotenv: '~6.1.0',
-      chalk: '~1.1.3',
-      sequelize: '~5.15.1',
+      express: '~4.16.3',
       [`forest-express-${orm}`]: '^5.2.0',
+      morgan: '1.9.1',
```
VincentMolinie: The only thing added here is that line. I just reorder by alphabetical order

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Command Generate - Add logs to requests on the project generated.
```
rap2hpoutre: Maybe

```suggestion
- Server - Add server logs on generated projects.
```

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
...
+
+### Changed
+- Command Generate - Update comments in routes.
```
rap2hpoutre: I think this is not required since it's an internal change of v3.

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
...
+
 ### Fixed
 - Command Generate - Fix smart routes for models composed of multiple words.
+- Command Generate - Rename Custom Actions to Smart Actions.
+- Command Generate - Rename Virtual to Smart.
+- Command Generate - Remove useless new lines in model and forest files.
```
rap2hpoutre: I think this is not required since it's an internal change of v3.

CHANGELOG.md
```diff
@@ -1,8 +1,18 @@
...
+- Technical - Drop APPLICATION_HOST in .env as it was useless.
```
```suggestion
- Technical - Drop `APPLICATION_HOST` in `.env` as it was useless.
```

CHANGELOG.md
VincentMolinie: Yeah but that was not released with the previous PR so I still think it should be here 😉

PR #336, opened 2019-11-21
[-] Templates - Replace unconventional `bin/www` script with `server.js`

rap2hpoutre:


services/dumper.js
```diff
@@ -64,7 +64,6 @@ function Dumper(config) {
       name: config.appName.replace(/ /g, '_').toLowerCase(),
       version: '0.0.1',
       private: true,
-      scripts: { start: 'node ./bin/www' },
```
VincentMolinie: [Spec] We cannot remove this as it will break the start with nodemon

rap2hpoutre: I used to run `nodemon bin/www` and now I have to run `nodemon server.js`. Was there another (better) way?

VincentMolinie: `nodemon` simply. `Nodemon` use the start by default ^^

rap2hpoutre: OK `nodemon` worked too.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- Templates - Relace unconventional `bin/www` script with `server.txt`.
```
VincentMolinie:
```suggestion
- Templates - Replace unconventional `bin/www` script with `server.txt`.
```

PR #333, opened 2019-11-20
[-] Technical - Use handlebars in lumber model generation

slimee:

PR #337, opened 2019-11-22
[*] MySQL - Handle BIT(1) boolean columns and handle the buffer<01> value returned on join queries.

SteveBunlon:


SteveBunlon:
[P3 - Boolean Fields - Using MySQL bit(1) fields as boolean breaks the UI](https://trello.com/c/mqJEFwP1/3861-p3-boolean-fields-using-mysql-bit1-fields-as-boolean-breaks-the-ui)

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Fixed
+- MySQL - Handle BIT(1) boolean columns and handle the buffer<01> value returned on join queries.
```
larcin: [Misc] Is this really a fix ? Looks like a change to me.

templates/app/models/index.js
```diff
@@ -18,6 +18,18 @@ let databaseOptions = {
...
 
+<%if(config.dbDialect === 'mysql') {
+%>databaseOptions.dialectOptions.typeCast = ( field, useDefaultTypeCasting ) => {
+  if ( ( field.type === "BIT" ) && ( field.length === 1 ) ) {
+    const bytes = field.buffer();
+    return( bytes[0] === 1 );
+  }
+
+  return( useDefaultTypeCasting() );
+};
```
larcin: [Code style] Why is there spaces in all those parenthesis ?

CHANGELOG.md
SteveBunlon: Not really actually. We are supposed to handle booleans on mysql since ever. And for some reason it doesn't work today. 

Moreover, the mysql boolean support was pretty bad so far, especially on queries using joins. 

IMO it should stay on the fixed section. What would be the conclusion to you ?

templates/app/models/index.js
SteveBunlon: FIxed

rap2hpoutre:
@SteveBunlon Thank you for your contribution. Before I review it, could you check with @VincentMolinie for the `BIT(1)` part? I suggested something similar to your change here: https://github.com/ForestAdmin/lumber/pull/309#pullrequestreview-309811893 and @VincentMolinie warned:

> This was removed for a good reason, this doesn't works on postgresql

So I guess his comment is still valid, not sure to understand the impact though.

templates/app/models/index.js
```diff
@@ -18,6 +18,18 @@ let databaseOptions = {
...
+  if ((field.type === "BIT") && (field.length === 1)) {
+    const bytes = field.buffer();
+    return bytes[0] === 1;
+  }
+
+  return useDefaultTypeCasting();
+};
+<%}
+%>
+
```
rap2hpoutre: I'm not sure, but it seems Lumber now uses Handlebars: https://github.com/ForestAdmin/lumber/pull/333.

SteveBunlon: Indeed you are right. It would be nice to rewrite the file and use handlebars instead. However I would rather prefer to limit the scope of this PR, and focus on the fix instead of rewriting the whole file.

test/expected/mysql/customers.json
```diff
@@ -21,6 +21,13 @@
       "primaryKey": false,
       "defaultValue": "1"
     },
+    {
+      "name": "paying",
+      "nameColumn": "paying",
+      "type": "BOOLEAN",
+      "primaryKey": false,
+      "defaultValue": "b'0'"
```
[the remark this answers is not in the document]
SteveBunlon: To avoid confusion, I changed the initial table script and switched the default value to 'null' instead.

SteveBunlon:
> Just found a bug (I don't exactly remember what I did but I think it's somehow related to the smart field value creation or edition):
> 
> Screenshot 2019-11-26 at 17 31 56

This was a bug indeed. I forgot to handle nullable boolean value. Now you can set boolean values to null (if the DB allows it obviously).

PR #338, opened 2019-11-28
[*] Remove useless files

slimee:


slimee:
Hi Adrien!

I think about you for this PR 😄 

I've removed two templates files that are not used anymore. 
These files are now found under the `models` folder.

To make sure this work, you can pull the repo and run tests.
As you can verify, the new files are well used in the tests now

See you!
🚀

PR #342, opened 2019-11-29
[+] Lumber - Add databases turnkeys sources.

slimee:
This PR is only adding the databases folder.
It contains sources for databases turnkeys.
SQL SERVER is excluded for now because the image doesn't works.


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 ### Added
+- Lumber - Add databases turnkeys sources.
```
VincentMolinie: It's look pretty cool. But I don't think it should be in that repository 🤔 ? It might be a new repo ? A repo where we can put all our initialisation script and these databases ^^

databases/mongo/fixtures/init.js
```diff
@@ -0,0 +1,86 @@
+/* eslint-disable */
+db.createCollection("films");
+db.createCollection("persons");
+
+
```
VincentMolinie: useless new line here

databases/mongo/fixtures/init.js
```diff
@@ -0,0 +1,86 @@
+/* eslint-disable */
```
VincentMolinie: Why disable eslint ? we can keep it even for this file

databases/mongo/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM mongo:bionic
+
+COPY fixtures /docker-entrypoint-initdb.d/
+
+ENV MONGO_INITDB_ROOT_USERNAME forest
+ENV MONGO_INITDB_ROOT_PASSWORD secret
+ENV MONGO_INITDB_DATABASE forest
```
VincentMolinie: We should not name the database forest as it will be really confusing with our database which is totally different ^^

databases/postgres/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM postgres:9.5.6
+
+COPY fixtures /docker-entrypoint-initdb.d/
+
+ENV POSTGRES_USER forest
+ENV POSTGRES_PASSWORD secret
+ENV POSTGRES_DB forest
```
VincentMolinie: Missing new line after this

databases/postgres/fixtures/reviews.sql
```diff
@@ -0,0 +1,12 @@
+DROP TABLE IF EXISTS reviews;
```
VincentMolinie: Why is it the only one with a drop if table exists 🤔 ?

CHANGELOG.md
slimee: yes, I think about this one: https://github.com/ForestAdmin/forest-examples

slimee:
Moved to https://github.com/ForestAdmin/forest-examples

PR #347, opened 2019-12-04
[*] Command Generate - Rename model `sessions` to `session` to avoid conflict with forest routes

VincentMolinie:
# How to reproduce
Create a model named `sessions`.
Create a new project with lumber.
Try to login to your liana.

# What has been done ?
I just rename the model to `session` without the s to avoid conflict with the `POST /forest/sessions`

### DIFF
--- a/services/dumper.js
+++ b/services/dumper.js
@@ -76,6 +76,13 @@ function Dumper(config) {
     return _.kebabCase(table);
   }
 
+  function getModelName(table) {
+    if (table === 'sessions') {
+      return 'session';
+    }
+    return _.camelCase(table);
+  }
+
   function writeDotGitIgnore() {
     const templatePath = `${__dirname}/../templates/app/gitignore`;
     const template = _.template(fs.readFileSync(templatePath, 'utf-8'));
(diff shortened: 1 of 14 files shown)


VincentMolinie:
[(?) P1 - Users can't log in if they have a model called sessions](https://trello.com/c/zSj11msK/3935-p1-users-cant-log-in-if-they-have-a-model-called-sessions)

services/dumper.js
```diff
@@ -76,6 +76,13 @@ function Dumper(config) {
     return _.kebabCase(table);
   }
 
+  function getModelName(table) {
+    if (table === 'sessions') {
+      return 'session';
+    }
```
arnaudbesnier: It does not look to be a long term solution against this kind of route conflicts issue.

services/dumper.js
```diff
@@ -76,6 +76,13 @@ function Dumper(config) {
     return _.kebabCase(table);
   }
 
+  function getModelName(table) {
+    if (table === 'sessions') {
```
larcin: [Notice] You should add a comment here to explain why thus edge case needs to be handled

services/dumper.js
VincentMolinie: What solution do you suggest ?

PR #356, opened 2019-12-10
[*] Technical - Fix linted deleted files issue

slimee:
This PR allows committing a file deletion.
Before that, `pre-commit-hook` linter was adding the deleted files in the `listModifiedFiles`

Here is an example of the issue. After deleting a file, the linter wanted to lint it.
Screenshot 2019-12-10 at 15 41 58

### DIFF
--- a/.eslint-bin/pre-commit-hook.js
+++ b/.eslint-bin/pre-commit-hook.js
@@ -12,7 +12,7 @@ function getFilesModified(callback) {
     }
 
     listFilesModified = status.files
-      .filter(change => change.index !== 'D')
+      .filter((change) => change.index !== 'D')
       .map((file) => file.path)
       .filter((file) => file.endsWith('.js'));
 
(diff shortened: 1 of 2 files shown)


.eslint-bin/pre-commit-hook.js
```diff
@@ -12,6 +12,7 @@ function getFilesModified(callback) {
     }
 
     listFilesModified = status.files
+      .filter(change => change.index !== 'D')
```
VincentMolinie: Actually it's even more than that, it would be nice to remove the file from the list file as it is useless now

slimee: ? It is what it do, isn't it?
The added line do: filter files, do not take deleted ones

VincentMolinie: Et les tests passent pas

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 - MySQL - Handle BIT(1) boolean columns and handle the buffer<01> value returned on join queries.
 - Technical - Fix unexpected test errors while running in the CI.
 - Technical - Fix database connection pool issue on tests.
+- Technical - Fix lint on deleted files issue.
```
VincentMolinie:
```suggestion
- Technical - Fix pre-commit hook when deleting a file.
```

PR #357, opened 2019-12-11
[-] Docker - Use `alpine` as base node image

thatguydan:
The nodejs alpine image is a fraction of the size of the full base image (75MB compared with 903MB). 

Further, according to this post (https://snyk.io/blog/top-ten-most-popular-docker-images-each-contain-at-least-30-vulnerabilities/) as at February of this year, the larger nodejs image contains hundreds of vulnerabilities, compared with 0 for the alpine image.

I haven't observed any unusual behaviour after this change.


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- Docker - Use `alpine` as base node image.
```
```suggestion
- Command Generate - Use `alpine` as base node Docker image for generated projects.
```

rap2hpoutre: @thatguydan Could you commit this suggestion? (I "do not have permission to push to this repository", Github says when I click on "commit suggestion", probably because it's on a fork).

PR #365, opened 2019-12-23
[-] Technical - Use Jest instead of Mocha for the test base

arnaudbesnier:


arnaudbesnier:
> 🙏 🚀 Thank you for this job :) it gives us a better codebase.
> Feel free to give me it back after the steve rebase 👍

Thanks! Steve's PR will be rebased on this commit (not the other way around).
I think we can merge this one ;)

PR #366, opened 2019-12-23
Command Generate - Ensure that generated project on MongoDB use the latest `mongoose` dependency version for security reasons

rap2hpoutre:
@arnaudbesnier I selected you as a reviewer for the following reasons:
 - We discussed about this on slack.
 - You made this commit: https://github.com/ForestAdmin/forest-express-mongoose/commit/9ee264ef8e1b6ae4103311d7b06b68581558535a.

Three things as been done (because it did not work without that):
 - Since the linter do not accept template-like file, I had to rename `templates/app/models/index.js` to  `templates/app/models/index.txt` (the commit can not pass without that). On a side note, the template system is currently messy: a mix of `txt` files, `js` files with invalid javascript (like this one) and handlebars code. We should improve that.
 - I had to add `useUnifiedTopology: true` in connect. Without this, there is a warning: https://mongoosejs.com/docs/deprecations.html#useunifiedtopology (we do not use any of the option that can generate a problem: `autoReconnect, reconnectTries, reconnectInterval`.
 - The change itself: upgrading to 5.8.2

Capture d’écran 2019-12-23 à 16 01 15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 ### Changed
-- Technical - Upgrade `mongoose` to the latest version in template.
+- Command Generate - Ensure that generated project on MongoDB use the latest `mongoose` dependency version for security reasons.
 
 ## RELEASE 3.1.0 - 2019-12-20
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- Technical - Upgrade `mongoose` to the latest version in template.
 
 ## RELEASE 3.1.0 - 2019-12-20
 ### Added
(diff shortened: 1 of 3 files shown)


rap2hpoutre:
[Update mongoose in template to avoid security error message](https://trello.com/c/TWeU87TV/3975-update-mongoose-in-template-to-avoid-security-error-message)

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+### Changed
+- Technical - Upgrade `mongoose` to the latest version in template.
```
arnaudbesnier: This is not a "Technical" change from my point of view.
It is about "Vulnerabilities"/"Security".

"Upgrade `mongoose` to the latest version in template" is something you can already read in the changelog, so I think we should explain the intent.

```suggestion
- Command Generate - Ensure that generated project on MongoDB use the latest `mongoose` dependency version for security reasons.
```

arnaudbesnier:
To Rebase to remove the `CHANGELOG.md` conflict.

PR #370, opened 2019-12-30
[*] Technical - Fix the missing dependencies in the `package.json` file of the generated project

arnaudbesnier:
I added the same package version as the `forest-express` package to prevent unexpected issues (as it was the forest-express dependency that was bringing the packages to the project).


arnaudbesnier:
[P3 - Lumber - "cors" and "express-jwt" dependencies are used by the generated project but not defined in the package.json](https://trello.com/c/RU50ylYK/3986-p3-lumber-cors-and-express-jwt-dependencies-are-used-by-the-generated-project-but-not-defined-in-the-packagejson)

PR #371, opened 2019-12-30
[*] Command Generate - Validate and fix the package name in `package.json` to prevent error on generated server startup

rap2hpoutre:
### Description

This fix avoids an error on project startup when project name is not allowed as a package name.

It uses the official [validate-npm-package-name](https://www.npmjs.com/package/validate-npm-package-name) created by NPM team to validate the project name. If the validation fails, it tries to transform the string to something that works (since slugify is not sufficient and I did not found a lib, I created a grumpy transformer that disallow almost everything when validation fails).

### Side note about tests
(see: https://forestadmin.slack.com/archives/GAZF5Q5RV/p1577712751015700)

I can not have created a unit test about the new created function in a `utils` directory because it's already used by tests' utils; migrating the existing system is not _that_ simple (see Slack discussion). We have to find a way to do that, but since we don't like to mix subject in one PR, I have to postpone this one. That's why I named the test file `test/utils-to-valid-package-name.test.js` (it should have been `test/utils/to-valid-package-name.test.js`). See this Trello card about this side note: https://trello.com/c/FaSVaFbY/2169-lumber-move-test-helpers-out-of-test-utils-folder

### DIFF
--- a/utils/to-valid-package-name.js
+++ b/utils/to-valid-package-name.js
@@ -1,0 +1,21 @@
+const validate = require('validate-npm-package-name');
+
+module.exports = function toValidPackageName(packageName) {
+  const { validForNewPackages, validForOldPackages } = validate(packageName);
+  const isValid = validForNewPackages && validForOldPackages;
+
+  if (!isValid) {
+    // NOTICE: hard slugify mode (disallow almost everything)
+    return packageName.toLowerCase()
+      // Remove all non "a-z", "0-9", "-" characters with hyphen.
+      .replace(/[^a-z0-9\\-]/g, '-')
+      // Remove hyphen sequence (> 1).
+      .replace(/-{2,}/g, '-')
+      // Remove leading and trailing hyphen.
+      .replace(/^-|-$/g, '')
+      // If empty, replace with my-lumber-project.
+      .replace(/^-*$/g, 'my-lumber-project');
+  }
+
+  return packageName;
+};
--- a/utils/to-valid-package-name.js
+++ b/utils/to-valid-package-name.js
@@ -2,8 +2,8 @@ const validate = require('validate-npm-package-name');
 
 module.exports = function toValidPackageName(packageName) {
   function isValid(name) {
-    const { validForNewPackages, validForOldPackages } = validate(name);
-    return validForNewPackages && validForOldPackages;
+    const { validForNewPackages } = validate(name);
+    return validForNewPackages;
   }
 
   if (!isValid(packageName)) {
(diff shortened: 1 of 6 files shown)


rap2hpoutre:
[P2 - Lumber - An "invalid" project name prevent the generated server to start.](https://trello.com/c/m52qh5Sm/3985-p2-lumber-an-invalid-project-name-prevent-the-generated-server-to-start)

utils/to-valid-package-name.js
```diff
@@ -0,0 +1,21 @@
+const validate = require('validate-npm-package-name');
+
+module.exports = function toValidPackageName(packageName) {
+  const { validForNewPackages, validForOldPackages } = validate(packageName);
+  const isValid = validForNewPackages && validForOldPackages;
```
slimee: Why only `validForNewPackages` check is not enough?

utils/to-valid-package-name.js
```diff
@@ -0,0 +1,21 @@
...
+    return packageName.toLowerCase()
+      // Remove all non "a-z", "0-9", "-" characters with hyphen.
+      .replace(/[^a-z0-9\\-]/g, '-')
+      // Remove hyphen sequence (> 1).
+      .replace(/-{2,}/g, '-')
+      // Remove leading and trailing hyphen.
+      .replace(/^-|-$/g, '')
+      // If empty, replace with my-lumber-project.
+      .replace(/^-*$/g, 'my-lumber-project');
+  }
```
slimee: We are not sure the returned string is valid, right? 
If the warranty is the replacement, then we don't need a validate lib.

Why not 1/replace then 2/check validity and 3/ throw if it's not valid?

slimee: Or maybe always return replacement and remove the lib?

utils/to-valid-package-name.js
rap2hpoutre: Yes `validForNewPackages` is enough, thank you! `validForOldPackages` seems to be for legacy purpose only, and is more permissive, so that's not what we want! 

Fixed!

utils/to-valid-package-name.js
rap2hpoutre: > maybe always return replacement and remove the lib?

I think we can keep the actual name without a drastic sanitization when it is valid: so it is returned as is if it's valid.

> We are not sure the returned string is valid, right?

Yes you are right. The new process is:

1. Validate it (if valid return it as is)
2. Sanitize it.
4. Validate it again (if valid return it, if not valid return `lumber-project`)

PR #372, opened 2019-12-31
[*] Command Generate - Do not generate route file for `sessions` model to avoid conflict with internal route

VincentMolinie:


VincentMolinie:
[(?) P1 - Users can't log in if they have a model called sessions](https://trello.com/c/zSj11msK/3935-p1-users-cant-log-in-if-they-have-a-model-called-sessions)

PR #378, opened 2020-01-06
[+] Technical - Test on postgres from 9.4 to 12.1

slimee:


slimee:
[(?) P1 - Lumber - Generate command fails on a MySQL v5.6 database](https://trello.com/c/r3SxgTi4/3988-p1-lumber-generate-command-fails-on-a-mysql-v56-database)

test-utils/database-urls.js
```diff
@@ -4,4 +4,5 @@ module.exports = {
   DATABASE_URL_MSSQL: 'mssql://sa:forest2019:@localhost:1432/model',
   DATABASE_URL_MYSQL: 'mysql://forest:secret@localhost:8999/lumber-sequelize-test',
   DATABASE_URL_POSTGRESQL: 'postgres://forest:secret@localhost:54369/lumber-sequelize-test',
+  DATABASE_URL_POSTGRESQL_OLD: 'postgres://forest:secret@localhost:54368/lumber-sequelize-test',
```
VincentMolinie: The only comment I will leave is here, not sure we should name it old as if we want to add another version it will be harder to distinguish them. May be we should add the version number in the variable name

arnaudbesnier: We could follow the Mongodb naming convention above ;)

docker-compose.yml
```diff
@@ -12,9 +12,19 @@ MongoDB:
...
+  image : postgres:9.4
+  container_name : lumber_sequelize_postgres_9_4
+  ports :
+    - "54368:5432"
+  environment:
+    - POSTGRES_DB=lumber-sequelize-test
+    - POSTGRES_USER=forest
+    - POSTGRES_PASSWORD=secret
+
+postgres121 :
```
arnaudbesnier: [Maintenance] `OldPostgres :` vs `postgres121` is not really consistent.

docker-compose.yml
```diff
@@ -12,9 +12,19 @@ MongoDB:
   volumes:
     - ./data/mongodb/4.2:/var/lib/mongodb
 
-postgres :
-  image : postgres:9.4.5
-  container_name : lumber_sequelize_postgres
+OldPostgres :
```
arnaudbesnier:
```suggestion
OldPostgres:
```

docker-compose.yml
arnaudbesnier:
```suggestion
postgres121:
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 ### Added
+- Technical - Test on postgres from 9.4 to 12.1.
```
arnaudbesnier: [Conventions] By convention we accumulate changes by the bottom.
Conventional commit will come to the rescue.

test-utils/database-urls.js
slimee: I understand your opinion guys. I started with version numbers in this file :) 
But after some usage, I found this was not useful, and maybe error-prone since this file can be not in sync with `docker-compose` and with the `multiple-database-version-helper.js::sqlDatabases`.

If it sounds for you, we can change to have only _OLD and _LAST suffixes in this file, and the file won't change except adding a new kind of DB.

arnaudbesnier: OK, easier to maintain without version number in the name.
But also less contextual when you read the variable name in the code.

Whatever the choice for consistency we should not live with 2 "conventions" (`_3_2` vs `_OLD`), so MongoDB variable names should be impacted by this new decision.

What about this:
```javascript
module.exports = {
  DATABASE_URL_MONGODB_MIN: 'mongodb://localhost:27015',
  DATABASE_URL_MONGODB_MAX: 'mongodb://localhost:27016',
  DATABASE_URL_MSSQL_MAX: 'mssql://sa:forest2019:@localhost:1432/model',
  DATABASE_URL_MYSQL_MAX: 'mysql://forest:secret@localhost:8999/lumber-sequelize-test',
  DATABASE_URL_POSTGRESQL_MIN: 'postgres://forest:secret@localhost:54368/lumber-sequelize-test',
  DATABASE_URL_POSTGRESQL_MAX: 'postgres://forest:secret@localhost:54369/lumber-sequelize-test',
};

```

slimee: MIN/MAX looks better than OLD. Like that I won't think about my grandma at work :)

slimee: It deserves a deeper look, but we should delete this file, and delete `multiple-database-version-helper.js::sqlDatabases` variable too. 
We may generate `sqlDatabases` directly from reading docker-compose.yml
A good way to have nothing written in multiple places, isn't it?

PR #379, opened 2020-01-06
[+] Technical - Test on mssql 2017 and 2019

slimee:


slimee:
[(?) P1 - Lumber - Generate command fails on a MySQL v5.6 database](https://trello.com/c/r3SxgTi4/3988-lumber-generate-command-fails-on-a-mysql-v56-database)

rap2hpoutre:
Just before I review it carefully: what was the consensus on this PR: https://github.com/ForestAdmin/lumber/pull/378? "Mssql2017" seems better than "OldMssql" to me but I don't want to pollute with a pointless review if something else was decided elsewhere. Should we change to `MIN` and `MAX`?

slimee:
> Just before I review it carefully: what was the consensus on this PR: #378? "Mssql2017" seems better than "OldMssql" to me but I don't want to pollute with a pointless review if something else was decided elsewhere. Should we change to `MIN` and `MAX`?

Yes I will set it on MIN/MAX 👍

PR #385, opened 2020-01-16
[*] Models - Fix error in mongoose projects preventing models from being exported.

SteveBunlon:


SteveBunlon:
[P3 - Lumber - The current way of exporting / importing models is broken (mongoose)](https://trello.com/c/hrG67LrN/4033-p3-lumber-the-current-way-of-exporting-importing-models-is-broken-mongoose)

PR #386, opened 2020-01-17
[*] Models - Fix error in mongoose projects preventing models from being exported.

SteveBunlon:


SteveBunlon:
[P3 - Lumber - The current way of exporting / importing models is broken (mongoose)](https://trello.com/c/hrG67LrN/4033-p3-lumber-the-current-way-of-exporting-importing-models-is-broken-mongoose)

SteveBunlon:
[P3 - Lumber - The current way of exporting / importing models is broken (mongoose)](https://trello.com/c/hrG67LrN/4033-p3-lumber-the-current-way-of-exporting-importing-models-is-broken-mongoose)

templates/app/models/index.hbs
```diff
@@ -1,7 +1,18 @@
 {{#if isMongoDB}}
 const mongoose = require('mongoose');
+{{#each models as |model|}}
+const {{model.modelName}} = require('./{{model.modelFile}}');
+{{/each}}
```
arnaudbesnier: Is it really maintainable to have to require all model files here.
What if I want to add a new model definition? Will I have to go in this file to require it manually?

SteveBunlon: We decided to do as such to fit with the way we import / exports models in sequelize projects. All models are accessible through the index file everywhere in the app. 

I didn't find any clear way of exporting / importing every models with mongoose. So if you have any other idea, I take it.

PR #388, opened 2020-01-20
[*] Command generate - Fix project generation on table with number as name

arnaud-moncel:
Fix error on project generation with database contain table with number as name.

### DIFF
--- a/services/dumper.js
+++ b/services/dumper.js
@@ -163,11 +163,14 @@ function Dumper(config) {
       };
     });
 
+    const modelSafeName = stringUtils.transformToSafeString(table);
+
     copyHandleBarsTemplate({
       source: `app/models/${config.dbDialect === 'mongodb' ? 'mongo' : 'sequelize'}-model.hbs`,
       target: `models/${tableToFilename(table)}.js`,
       context: {
-        modelName: stringUtils.transformToSafeString(stringUtils.pascalCase(table)),
+        modelVariableName: stringUtils.pascalCase(modelSafeName),
+        modelName: stringUtils.pascalCase(modelSafeName, false),
         table,
         fields: fieldsDefinition,
         references: referencesDefinition,
@@ -186,8 +189,7 @@ function Dumper(config) {
       source: 'app/routes/route.hbs',
       target: `routes/${tableToFilename(modelName)}.js`,
       context: {
-        tableName: modelName,
-        modelName: stringUtils.transformToSafeString(stringUtils.pascalCase(modelName)),
+        modelName: stringUtils.pascalCase(stringUtils.transformToSafeString(modelName), false),
         modelNameReadablePlural: plural(readableModelName),
         modelNameReadableSingular: singular(readableModelName),
         isMongoDB: config.dbDialect === 'mongodb',
@@ -199,7 +201,10 @@ function Dumper(config) {
     copyHandleBarsTemplate({
       source: 'app/forest/collection.hbs',
       target: `forest/${tableToFilename(table)}.js`,
-      context: { isMongoDB: config.dbDialect === 'mongodb', table },
+      context: {
+        isMongoDB: config.dbDialect === 'mongodb',
+        table: stringUtils.pascalCase(stringUtils.transformToSafeString(table), false),
+      },
     });
   }
 
--- a/services/dumper.js
+++ b/services/dumper.js
@@ -167,7 +167,7 @@ function Dumper(config) {
       source: `app/models/${config.dbDialect === 'mongodb' ? 'mongo' : 'sequelize'}-model.hbs`,
       target: `models/${tableToFilename(table)}.js`,
       context: {
-        modelName: stringUtils.pascalCase(table),
+        modelName: stringUtils.transformToSafeString(stringUtils.pascalCase(table)),
         table,
         fields: fieldsDefinition,
         references: referencesDefinition,
@@ -180,16 +180,14 @@ function Dumper(config) {
   }
 
   function writeRoute(modelName) {
-    const modelNameDasherized = _.kebabCase(modelName);
     const readableModelName = _.startCase(modelName);
 
     copyHandleBarsTemplate({
       source: 'app/routes/route.hbs',
       target: `routes/${tableToFilename(modelName)}.js`,
       context: {
-        table: modelName,
-        modelName: stringUtils.pascalCase(modelName),
-        modelNameDasherized,
+        tableName: modelName,
+        modelName: stringUtils.transformToSafeString(stringUtils.pascalCase(modelName)),
         modelNameReadablePlural: plural(readableModelName),
         modelNameReadableSingular: singular(readableModelName),
         isMongoDB: config.dbDialect === 'mongodb',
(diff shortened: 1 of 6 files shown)


arnaud-moncel:
[P3 - Lumber generates a broken project if at least one table name is a "number"](https://trello.com/c/AsDIdaaD/29-p3-lumber-generates-a-broken-project-if-at-least-one-table-name-is-a-number)

utils/strings.js
```diff
@@ -2,6 +2,10 @@ const _ = require('lodash');
 
 module.exports = {
   pascalCase(input) {
+    if (/^[\d]/g.exec(input)) {
+      return `Table${input}`;
+    }
```
arnaudbesnier: [Consistency] Not convinced this is a good idea to place such specific code logic in a function named "pascalCase" and that now does not exactly "pascalCase" the input value.

services/dumper.js
```diff
@@ -187,7 +187,8 @@ function Dumper(config) {
       source: 'app/routes/route.hbs',
       target: `routes/${tableToFilename(modelName)}.js`,
       context: {
-        modelName,
+        table: modelName,
```
arnaudbesnier:
```suggestion
        tableName: modelName,
```
?

arnaudbesnier: Or 
```javascript
        modelName,
        modelNameSafe: stringUtils.pascalCase(modelName),
```
?

templates/app/routes/route.hbs
```diff
@@ -3,50 +3,50 @@ const { PermissionMiddlewareCreator } = require('forest-express-{{#if isMongoDB
...
+const permissionMiddlewareCreator = new PermissionMiddlewareCreator('{{ table }}');
 
-// This file contains the logic of every route in Forest Admin for the collection {{ modelName }}:
+// This file contains the logic of every route in Forest Admin for the collection {{ table }}:
 // - Native routes are already generated but can be extended/overriden - Learn how to extend a route here: https://docs.forestadmin.com/documentation/v/v5/reference-guide/routes/extend-a-route
 // - Smart action routes will need to be added as you create new Smart Actions - Learn how to create a Smart Action here: https://docs.forestadmin.com/documentation/v/v5/reference-guide/actions/create-and-manage-smart-actions
 
 // Create a {{ modelNameReadableSingular }}
-router.post('/{{ modelName }}', permissionMiddlewareCreator.create(), (request, response, next) => {
+router.post('/{{ modelNameDasherized }}', permissionMiddlewareCreator.create(), (request, response, next) => {
```
arnaudbesnier: Is it part of the issue?
Why should we have to change the route names?

arnaud-moncel: No, my bad. I revert this.

utils/strings.js
```diff
@@ -1,7 +1,18 @@
 const _ = require('lodash');
 
 module.exports = {
-  pascalCase(input) {
-    return _.chain(input).camelCase().upperFirst().value();
+  pascalCase(input, upperFirst = true) {
```
arnaudbesnier: This is "PascalCase" and this is "camelCase".
So we should not expose a method that is named `pascalCase` but has an argument to create generate camel cased strings.

```javascript
pascalCase(input) {
  return _.chain(input).camelCase().upperFirst().value();
}

camelCase(input) {
   return _.camelCase(value);
}
```
is much better in my opinion.

services/dumper.js
```diff
@@ -163,11 +163,14 @@ function Dumper(config) {
...
     });
 
+    const modelSafeName = stringUtils.transformToSafeString(table);
+
     copyHandleBarsTemplate({
       source: `app/models/${config.dbDialect === 'mongodb' ? 'mongo' : 'sequelize'}-model.hbs`,
       target: `models/${tableToFilename(table)}.js`,
       context: {
-        modelName: stringUtils.pascalCase(table),
+        modelVariableName: stringUtils.pascalCase(modelSafeName),
```
arnaudbesnier: I don't understand what is the purpose of this new contextual variable.
From my understanding it adds more complexity (and diff in this PR) for a value that I don't get.
Can't we keep using the `modelName` in the template?

services/dumper.js
```diff
@@ -180,15 +183,13 @@ function Dumper(config) {
   }
 
   function writeRoute(modelName) {
-    const modelNameDasherized = _.kebabCase(modelName);
```
arnaudbesnier: Is it a cleaning of the existing code?
If so it should not be present in this PR to make it as easy as possible to read.

I am a bit picky here, but I think we should do that to make the best collaboration/review possible.
Every tiny details could be noise for the reviewer.

(See Pull Requests - Code of Conduct: https://forest.slite.com/app/channels/-KdBU14ToWBAerCk_HGS/notes/QFFoNQ1pOg#68229e06)

services/dumper.js
```diff
@@ -200,7 +201,10 @@ function Dumper(config) {
     copyHandleBarsTemplate({
       source: 'app/forest/collection.hbs',
       target: `forest/${tableToFilename(table)}.js`,
-      context: { isMongoDB: config.dbDialect === 'mongodb', table },
+      context: {
+        isMongoDB: config.dbDialect === 'mongodb',
+        table: stringUtils.pascalCase(stringUtils.transformToSafeString(table), false),
```
arnaudbesnier: As this pattern is present several time, you should definitely now consider a "business oriented" formatter that uses technical formatters:
```javascript
function getModelNameFromTableName(table) {
  return stringUtils.pascalCase(stringUtils.transformToSafeString(table), false);
}
```
It will definitely ease the future maintenance of this logic.

utils/strings.js
arnaud-moncel: oh yes my bad.

services/dumper.js
arnaud-moncel: I thinked we wanted
```js
const Model = sequelize.define('model', {
```

take this example :
the table name is `1`
the safe model name is `model1`

do you prefer (which is the current solution)
```js
const Model1 = sequelize.define('model1', {
```

or

```js
const model1 = sequelize.define('model1', {
```

services/dumper.js
arnaud-moncel: Sorry about that, you are right ! it's difficult to loose bad practice 🤣

services/dumper.js
arnaudbesnier: ok, yes we want:
`const Model1 = sequelize.define('model1', {`
Forget my comment. I didn't get it because of the "refactoring" included in the change:
`modelName` renamed in `modelVariableName`.

services/dumper.js
```diff
@@ -163,11 +163,14 @@ function Dumper(config) {
...
 
+    const modelSafeName = stringUtils.transformToSafeString(table);
+
     copyHandleBarsTemplate({
       source: `app/models/${config.dbDialect === 'mongodb' ? 'mongo' : 'sequelize'}-model.hbs`,
       target: `models/${tableToFilename(table)}.js`,
       context: {
-        modelName: stringUtils.pascalCase(table),
+        modelVariableName: stringUtils.pascalCase(modelSafeName),
+        modelName: stringUtils.camelCase(modelSafeName),
```
arnaudbesnier: My last suggestion
```suggestion
        modelName: getModelNameFromTableName(table),
```
Yes it is certainly less performant in terms of code execution but much more maintainable because we understand the the intent using `getModelNameFromTableName` versus the current code.

PR #389, opened 2020-01-22
[-] Models - Improve models exports to support model addition / deletion without requiring changes.

SteveBunlon:
For this PR I would suggest to not change the changelog, and to squash the present work with this one : https://github.com/ForestAdmin/lumber/pull/386

As the subjects are related and next to each other in the git history, It would be much cleaner and tinier.

### DIFF
--- a/templates/app/models/index.hbs
+++ b/templates/app/models/index.hbs
@@ -4,7 +4,7 @@ const requireAll = require('require-all');
 
 const models = {};
 
-mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true });
+mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true, useUnifiedTopology: true });
 
 function filterFile(filename) {
   if (filename.endsWith('.js') && filename !== 'index.js') {
--- a/templates/app/models/index.hbs
+++ b/templates/app/models/index.hbs
@@ -4,7 +4,7 @@ const requireAll = require('require-all');
 
 const models = {};
 
-mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true });
+mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true, useUnifiedTopology: true });
 
 const modelsList = requireAll({
   dirname:  __dirname,
--- a/templates/app/models/index.hbs
+++ b/templates/app/models/index.hbs
@@ -7,7 +7,7 @@ const models = {};
 mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true, useUnifiedTopology: true });
 
 const modelsList = requireAll({
-  dirname:  __dirname,
+  dirname: __dirname,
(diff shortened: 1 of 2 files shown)


SteveBunlon:
[P3 - Lumber - The current way of exporting / importing models is broken (mongoose)](https://trello.com/c/hrG67LrN/131-p3-lumber-the-current-way-of-exporting-importing-models-is-broken-mongoose)

templates/app/models/index.hbs
```diff
@@ -1,16 +1,21 @@
...
 
-const models = {
-{{#each models as |model|}}
-  {{model.modelName}},
-{{/each}}
-};
+mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true });
+
+const modelsList = requireAll({
+  dirname:  __dirname,
```
arnaudbesnier:
```suggestion
  dirname: __dirname,
```
?

templates/app/models/index.hbs
```diff
@@ -1,16 +1,21 @@
...
+  filter: `^(?!index\.js$).+`,
+  resolve: (Model) => Model,
```
arnaudbesnier: Maybe `recursive: true` would be a good addition for more flexible code in the future.

templates/app/models/index.hbs
```diff
@@ -1,16 +1,21 @@
...
-const models = {
-{{#each models as |model|}}
-  {{model.modelName}},
-{{/each}}
-};
+mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true });
+
+const modelsList = requireAll({
+  dirname:  __dirname,
+  filter: `^(?!index\.js$).+`,
```
arnaudbesnier: Following what we have in `forest-express` `requireAll`, maybe:
```javascript
  excludeDirs: /^__tests__$/,
  filter: (fileName) => isJavascriptFileName(fileName) && !isTestFileName(fileName),
```
could be a good improvement.

templates/app/models/index.hbs
```diff
@@ -1,16 +1,21 @@
 {{#if isMongoDB}}
 const mongoose = require('mongoose');
-{{#each models as |model|}}
-const {{model.modelName}} = require('./{{model.modelFile}}');
-{{/each}}
+const requireAll = require('require-all');
 
-mongoose.connect(process.env.DATABASE_URL, { useNewUrlParser: true, useUnifiedTopology: true });
```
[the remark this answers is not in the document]
SteveBunlon: Wrong copy past i guess, this is not needed, I added it back.

templates/app/models/index.hbs
SteveBunlon: As discussed together, this will be handled in the future.

templates/app/models/index.hbs
SteveBunlon: As discussed together, this will be handled in the future.

PR #394, opened 2020-02-11
[*] Command generate - Fix error on project generation when tables have references on table with non safe name

arnaud-moncel:


rap2hpoutre:
See: https://github.com/ForestAdmin/lumber/pull/388

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Fixed
 - Command Generate - Fix error on project generation with database contain table with number as name.
+- Command Generate - Fix error on project generation when tabbles have references on table with non safe name.
```
```suggestion
- Command Generate - Fix error on project generation when tables have references on table with non safe name.
```

services/dumper.js
```diff
@@ -144,7 +144,11 @@ function Dumper(config) {
       const expectedConventionalColumnName = underscored ? _.snakeCase(field.name) : field.name;
       const nameColumnUnconventional = field.nameColumn !== expectedConventionalColumnName
         || (underscored && /[1-9]/g.test(field.name));
-      return { ...field, nameColumnUnconventional };
+      let ref;
+      if (field.ref) {
+        ref = getModelNameFromTableName(field.ref);
+      }
+      return { ...field, ref, nameColumnUnconventional };
```
rap2hpoutre: Nice job, you haven't mutated the original field! On a side note, just because I prefer anything over `let` (but you could consider this as a kind of madness), I would have preferred something like:

```javascript
return {
  ...field,
  ref: field.ref && getModelNameFromTableName(field.ref), // or a ternary operator
  nameColumnUnconventional,
};
```

But it is an opinion (I may be wrong) that you are free to ignore! (I just wanted to leave a comment in your PR)

PR #403, opened 2020-03-17
Command Generate - Export ID fields and force isReadOnly property where needed

guillaumedeslandes:
This adds information, but `isReadOnly: false` is not honoured by frontend.


templates/app/models/sequelize-model.hbs
```diff
@@ -11,6 +11,7 @@ module.exports = (sequelize, DataTypes) => {
       field: '{{field.nameColumn}}',{{/if}}{{#if field.primaryKey}}
       primaryKey: true,{{/if}}{{#if field.defaultValue}}
       defaultValue: Sequelize.literal('{{{field.defaultValue}}}'),{{/if}}
+      isReadOnly: {{field.isReadOnly}},
```
arnaudbesnier: Is `isReadOnly` a Sequelize field attribute?
https://sequelize.org/master/manual/model-basics.html#concept

PR #407, opened 2020-03-23
[*] Documentation - Update documentation links to v6

rap2hpoutre:
Lint fails, it is fixed here: https://github.com/ForestAdmin/lumber/pull/408


templates/app/models/sequelize-model.hbs
```diff
@@ -20,7 +20,7 @@ module.exports = (sequelize, DataTypes) => {
     schema: process.env.DATABASE_SCHEMA,{{/if}}
   });
 {{#if noId}}  {{modelVariableName}}.removeAttribute('id');{{/if}}
-  // This section contains the relationships for this model. See: https://docs.forestadmin.com/documentation/v/v5/reference-guide/relationships#adding-relationships.
+  // This section contains the relationships for this model. See: https://docs.forestadmin.com/documentation/v/v6/reference-guide/relationships#adding-relationships.
```
jeffladiray: Pretty sure this is not related to your PR, but most of the lines doesn't end with a final dot

rap2hpoutre:
> Minor suggestion - isn't that a pain to manually update, on each release, all thoses templates files ?

I guess it's ≈OK for now.

PR #409, opened 2020-03-24
[+] Command Generate - Prevent error when table name is a JS reserved word (#3upt3a)

guillaumedeslandes:


utils/strings.js
```diff
@@ -1,5 +1,24 @@
 const _ = require('lodash');
 
+const reservedWords = [
```
arnaud-moncel: I think we can add `module` no ?
what happened if 
```js
const { module } = require('model');

module.exports = { };
```

guillaumedeslandes: will do

larcin: [Code Style] I think that we tend to have hard coded const like that capitalized and underscored:
```suggestion
const RESERVED_WORDS = [
```

utils/strings.js
```diff
@@ -9,10 +28,18 @@ module.exports = {
...
 
+  isReservedWord(input) {
+    return reservedWords.includes(_.toLower(input));
+  },
+
   transformToSafeString(input) {
     if (/^[\d]/g.exec(input)) {
       return `model${input}`;
     }
+    // Notice: add dash to get proper snake/pascal case
```
larcin:
```suggestion
    // NOTICE: add dash to get proper snake/pascal case
```

utils/strings.js
```diff
@@ -9,10 +28,18 @@ module.exports = {
     return _.camelCase(input);
   },
 
+  isReservedWord(input) {
+    return reservedWords.includes(_.toLower(input));
```
larcin:
```suggestion
    return RESERVED_WORDS.includes(_.toLower(input));
```

PR #410, opened 2020-03-25
[*] Command Generate - Remove extraneous calls to Sequelize.literal in generated models

guillaumedeslandes:
Generated model files used Sequelize.literal for all default value. This is not necessary (as per [ForestAdmin doc](https://docs.forestadmin.com/documentation/reference-guide/models/enrich-your-models#adding-a-default-value-to-your-models) and lead to simple values being displayed as `[object Object]` in ForestAdmin create forms.

Goes with https://github.com/ForestAdmin/forest-express-sequelize/pull/389


test/services/dumper/dumper-sequelize.test.js
```diff
@@ -72,4 +73,15 @@ describe('services > dumper > sequelize', () => {
     expect(generatedRouteFile).toStrictEqual(expectedRouteFile);
     cleanOutput();
   });
+
+  it('should generate a model with defaul values', async () => {
```
rap2hpoutre:
```suggestion
  it('should generate a model with default values', async () => {
```

services/dumper.js
```diff
@@ -14,6 +15,13 @@ const mkdirp = P.promisify(mkdirpSync);
 
 const DEFAULT_PORT = 3310;
 
+const DEFAULT_VALUE_TYPES_TO_STRINGIFY = [
+  `${Sequelize.DataTypes.ARRAY}`,
+  `${Sequelize.DataTypes.ENUM}`,
+  `${Sequelize.DataTypes.JSONB}`,
+  `${Sequelize.DataTypes.STRING}`,
```
rap2hpoutre: Should we consider `TEXT` too? (not sure)

services/dumper.js
```diff
@@ -137,18 +145,42 @@ function Dumper(config) {
...
+    let safeDefaultValue = field.defaultValue;
+    if (config.dbDialect !== 'mongodb') {
+      if (typeof safeDefaultValue === 'object' && safeDefaultValue instanceof Sequelize.Utils.Literal) {
+        safeDefaultValue = `Sequelize.literal('${safeDefaultValue.val}')`;
+      } else if (
+        !_.isNil(safeDefaultValue)
+        && _.some(
+          DEFAULT_VALUE_TYPES_TO_STRINGIFY,
+          // NOTICE: Uses `startsWith` as composite types may vary (eg: `ARRAY(DataTypes.INTEGER)`)
+          (dataType) => _.startsWith(field.type, dataType),
```
rap2hpoutre: [Thinking] This could be considered not reliable to analyze the beginning of the string (i.e `DATEONLY` vs `DATE`). However, in that particular case, we could argue it's not a problem! (but it could be later since the contract is hidden).

test/services/dumper/dumper-sequelize.test.js
```diff
@@ -5,6 +5,7 @@ const simpleModel = require('../../../test-expected/sequelize/db-analysis-output
 const belongsToModel = require('../../../test-expected/sequelize/db-analysis-output/addresses.expected.json');
 const otherAssociationsModel = require('../../../test-expected/sequelize/db-analysis-output/users.expected.json');
 const exportModel = require('../../../test-expected/sequelize/db-analysis-output/export.expected.json');
+const defaultValuesModel = require('../../../test-expected/sequelize/db-analysis-output/default-values.expected.js');
```
rap2hpoutre: This looks like a fixture more than an expected value. However, I see that this whole file use the same pattern, so I guess it's OK.

rap2hpoutre:
On other hand, I tested with this table and it works like a charm! 👌 Good job! 🎉 

```sql
create table tabl
(
	id serial not null
		constraint tabl_pk
			primary key,
	des varchar,
	q integer[] default '{1,2,3}'::integer[],
	column_4 integer default 123,
	column_5 varchar default 'fdfd'::character varying,
	column_6 date default now(),
	price numeric default 9.99,
	updt date default now(),
	column_9 jsonb default '[{"a": "b"}]'::jsonb
);

alter table tabl owner to forest;


```

services/dumper.js
guillaumedeslandes: Right.
Added `TEXT` and `CITEXT`.
https://sequelize.org/master/manual/model-basics.html#data-types

services/dumper.js
guillaumedeslandes: Agreed, but as you said, this should not be an issue here. Starting with `DATE` means "I am a date value", and the code says "All date values should be stringified". We should be safe for now.

Added `DATE` to the list of type to quote btw. Thanks!

test/services/dumper/dumper-sequelize.test.js
guillaumedeslandes: I had to use a JS file to because some expressions where needed (eg: `"defaultValue": Sequelize.literal('now()')`). Could not make this work with a fixture / JSON file.
Definitely used a fixture as the code base for this.

PR #411, opened 2020-04-01
[*] Command Generate - Fix issue where a conflict between regular fields name and references alias could happen, preventing the server from starting.

SteveBunlon:
clickup card --> https://app.clickup.com/t/3jf395

Using the generate command, sometimes we can have relationships named the same as regular fields. This causes a generated sequelize model to have two fields with the same name, and sequelize throws an error for this, preventing the server from starting. 

Here is an example of a situation where a field and a reference field alias would have the same name :
```
CREATE TABLE dept_manager (
   emp_no       INT             NOT NULL,
   dept_no      CHAR(4)         NOT NULL,
   from_date    DATE            NOT NULL,
   to_date      DATE            NOT NULL,
   FOREIGN KEY (emp_no)  REFERENCES employees (emp_no)    ON DELETE CASCADE,
   FOREIGN KEY (dept_no) REFERENCES departments (dept_no) ON DELETE CASCADE,
   PRIMARY KEY (emp_no,dept_no)
); 
```

Here lumber will generate a field named `emp_no` but also a reference to `employee` with the alias `emp_no`. 

To prevent this, we perform a conflict detection at the end of the DB scan process, and solve conflicts if any by suffixing the alias name with `Relationship`. 

To test this out, you can use the mysql sample DB from [here](https://github.com/datacharmer/test_db) and try to generate a project using the `devel` branch (you will see that the server startup fails). 

Using this branch it should fix the issue.


services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -299,6 +299,26 @@ async function createTableSchema({
   };
 }
 
+function detectAndFixAliasConflicts(wholeSchema) {
```
rap2hpoutre, line 302: - As _detecting_ is a part of _fixing_, we could consider renaming this function to `fixAliasConflicts` (maybe it could then be more obvious that it has side effects since the first word is `fix`).
- We could have added a notice to explain what specific case we fix here. Reading the code (without digging), I can't see at first glance that the problem is: "when a field as the same name as a relationship, we fix the relationship name". The changelog entry is explicit, maybe we could use something similar (disclaimer: I love comments).
 - (side note) I tried to re-implement this function in a more functional way (10 minutes), but I failed (it was less expressive and understandable), I prefer yours!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ### Fixed
 - Technical - Fix test expect files extensions.
 - Command Generate - Do not declare the "id" field on tables having a `BIGINT` "id" column as primary key. 🛡
+- Command Generate - Fix issue where a conflict between regular fields and references alias could happen, and prevent the server from starting.
```
rap2hpoutre, line 13: [Suggestion] 
```suggestion
- Command Generate - Fix issue where a conflict between regular fields and references alias could prevent the server from starting. 
```

Not sure though. Maybe yours is better

SteveBunlon: 👍

arnaudbesnier, line 13:
```suggestion
- Command Generate - Fix issue where a conflict between regular fields and references alias could happen, and prevent the server from starting. 🛡
```

services/analyzer/sequelize-tables-analyzer.js
SteveBunlon: I updated the name and added a notice 👍

CHANGELOG.md
SteveBunlon: 👍

PR #419, opened 2020-04-28
fix(generate): fix database url encoded in .env

VincentMolinie:
# Description
When installing a project with a mongo project with an authSource, the `=` get encoded

# Actual behaviour
I get a DATABASE_URL=mongo.....?authSource&#x3D;admin
So the server does not start

# Expected behaviour
To not have an encoded DATABASE_URL

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 - Technical - Adapt release manager to conventional commits.
 - Documentation - Update links in readme.
 
+### Fixed
+- Command Generate - Fix DATABASE_URL badly encoded in .env generation.
+
 ## RELEASE 3.5.1 - 2020-04-06
 ### Fixed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 - Technical - Adapt release manager to conventional commits.
 - Documentation - Update links in readme.
 
+### Fixed
+- Command Generate - Fix DATABASE_URL badly encoded in .env generation.
```
rap2hpoutre, line 11: The whole content is not encoded anymore by Handlebars, not just `DATABASE_URL`. I don't have a suggestion though (I don't like the one I suggested).

```suggestion
- Command Generate - Fix bad encoding of content in generated files (remove HTML escaping in template file).
```

PR #424, opened 2020-05-05
feat(command generate): support 'JSON' field type on installation

SteveBunlon:
## Pull Request checklist:

- [x] Write changes made in the CHANGELOG.md
- [x] Create automatic tests

[click up](https://github.com/ForestAdmin/lumber/pull/424)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Command Generate - Support `JSON` field type on installation.
+
 ### Changed
 - Technical - Add a Pull Requests template.
 - Technical - Remove CI configuration warnings.
--- a/services/analyzer/sequelize-column-type-getter.js
+++ b/services/analyzer/sequelize-column-type-getter.js
@@ -3,6 +3,7 @@ const logger = require('../logger');
 
 const DIALECT_MYSQL = 'mysql';
 const DIALECT_POSTGRES = 'postgres';
+const DIALECT_MSSQL = 'mssql';
 
 function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
   const queryInterface = databaseConnection.getQueryInterface();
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Change Log
 
 ## [Unreleased]
+### Added
+- Command Generate - Support `JSON` field type on installation.
```
rap2hpoutre, line 5: [Ignorable] Not sure about "on installation"...

```suggestion
- Command Generate - Support `JSON` field type.
```

services/analyzer/sequelize-column-type-getter.js
```diff
@@ -72,11 +73,14 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
     return `ARRAY(DataTypes.${await this.perform(innerColumnInfo, innerColumnInfo.udtName, tableName)})`;
   };
 
+  // eslint-disable-next-line sonarjs/cognitive-complexity
```
rap2hpoutre, line 76: I fell like we should try to avoid introducing cognitive complexity. The complexity was already there, you added one line, so it's not your responsibility. We also should not mix refactor and feature introduction. 

**But**, I think we could prefer a refactoring task over adding `eslint-disable-next-line`. WDYT?

### Suggestion

A small refactor could be replacing all these occurrence (they are considered as "complex" by the linter, we could consider these as duplicate):

```js
case type.match(/^SMALLINT.*/i) || {}).input:
```

With:
```js
case typeStartingWith(type, 'SMALLINT'):
```

Then create a function:
```js
function typeStartingWith(type, value) {
  return type.match(new RegExp(`^${value}.*`, 'i') || {}).input;
}
```

I just checked (I did not tested it though), it's enough to reduce the complexity detected by sonarJS. It also removes 11 duplicates lines. Still, I'm aware this is not related to this actual issue. Feel free to suggest something else (however, I feel like we have to do something to address this `eslint-disable` issue).

services/analyzer/sequelize-column-type-getter.js
```diff
@@ -72,11 +73,14 @@ function ColumnTypeGetter(databaseConnection, schema, allowWarning = true) {
...
   this.perform = async (columnInfo, columnName, tableName) => {
     const { type } = columnInfo;
     const mysqlEnumRegex = /ENUM\((.*)\)/i;
 
     switch (type) {
+      case (type === 'JSON' && !isDialect(DIALECT_MSSQL) && 'JSON'):
+        return 'JSON';
```
rap2hpoutre, line 83: [Ignorable] This syntax id not that common. However, I have no suggestion that does not require a full refactor (you mentioned the strategy pattern, but I guess it's out of scope).

arnaudbesnier:
Sorry, I destroyed the `devel` branch (for semantic-releases), I didn't expect GitHub to close the PRs. Can you re-create it based on `master` branch (without changelog line)?

SteveBunlon:
> Sorry, I destroyed the `devel` branch (for semantic-releases), I didn't expect GitHub to close the PRs. Can you re-create it based on `master` branch (without changelog line)?

👍

PR #445, opened 2020-06-28
fix: model generation inconsistencies

closingin:
Fixes [#4embh7](https://app.clickup.com/t/4embh7).

This PR contains to bug fixes in order to correctly handle snake_cased tables that only contain foreign keys.

The issue came from two different bugs in the code :
- The `through` property that is generated when having `belongsToMany` relationships was using the table name as its value, but it's meant to contain the name of the model, which is camelCased in our setup.
- The `options.underscored` property was only computed from the fields names, so if you had no fields but had some snake_cased FKs, it still defaulted to false.

The consequence was that in the schema building step, after having generated your project, Sequelize couldn't find the snake_cased model in the `through` property, and as such created it on its own (which is its normal behaviour) with default values (such as createdAt and updatedAt).

From the documentation :
> ```js
> Movie.belongsToMany(Actor, { through: 'ActorMovies' });
> Actor.belongsToMany(Movie, { through: 'ActorMovies' });
> ```
> Since a string was given in the through option of the belongsToMany call, Sequelize will automatically create the ActorMovies model which will act as the junction model. For example, in PostgreSQL:
>
> ```sql
> CREATE TABLE IF NOT EXISTS "ActorMovies" (
>  "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL,
>  "updatedAt" TIMESTAMP WITH TIME ZONE NOT NULL,
>  "MovieId" INTEGER REFERENCES "Movies" ("id") ON DELETE CASCADE ON UPDATE CASCADE,
>  "ActorId" INTEGER REFERENCES "Actors" ("id") ON DELETE CASCADE ON UPDATE CASCADE,
>  PRIMARY KEY ("MovieId","ActorId")
>);
>```

## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/services/analyzer/sequelize-tables-analyzer.js
+++ b/services/analyzer/sequelize-tables-analyzer.js
@@ -371,6 +371,12 @@ async function analyzeSequelizeTables(databaseConnection, config, allowWarning) 
   const referencesPerTable = createAllReferences(databaseSchema, schemaAllTables);
   Object.keys(referencesPerTable).forEach((tableName) => {
     schemaAllTables[tableName].references = _.sortBy(referencesPerTable[tableName], 'association');
+
+    if (!schemaAllTables[tableName].fields.length) {
+      schemaAllTables[tableName].options.underscored = isUnderscored(
+        schemaAllTables[tableName].references.map(({ foreignKey }) => ({ nameColumn: foreignKey })),
+      );
+    }
   });
 
   if (_.isEmpty(schemaAllTables)) {
--- a/services/analyzer/sequelize-tables-analyzer.js
+++ b/services/analyzer/sequelize-tables-analyzer.js
@@ -372,6 +372,9 @@ async function analyzeSequelizeTables(databaseConnection, config, allowWarning) 
   Object.keys(referencesPerTable).forEach((tableName) => {
     schemaAllTables[tableName].references = _.sortBy(referencesPerTable[tableName], 'association');
 
+    // NOTE: When a table contains no field, it will be considered camelCased
+    //       by default, so we need to check its references to ensure whether
+    //       it is camelCased or not.
     if (!schemaAllTables[tableName].fields.length) {
       schemaAllTables[tableName].options.underscored = isUnderscored(
         schemaAllTables[tableName].references.map(({ foreignKey }) => ({ nameColumn: foreignKey })),
--- a/services/dumper.js
+++ b/services/dumper.js
@@ -210,8 +210,12 @@ function Dumper(config) {
       }
 
       if (reference.targetKey) {
+        const expectedConventionalTargetKeyName = underscored
+          ? _.snakeCase(reference.targetKey)
+          : _.camelCase(reference.targetKey);
+
         computedReference.targetKeyColumnUnconventional = reference.targetKey
-          !== (underscored ? _.snakeCase(reference.targetKey) : _.camelCase(reference.targetKey));
+          !== expectedConventionalTargetKeyName;
       }
 
       return computedReference;
--- a/services/dumper.js
+++ b/services/dumper.js
@@ -200,23 +200,21 @@ function Dumper(config) {
     });
 
     const referencesDefinition = references.map((reference) => {
-      const isBelongsToMany = reference.association === 'belongsToMany';
+      const computedReference = {
+        ...reference,
+        isBelongsToMany: reference.association === 'belongsToMany',
+      };
+
+      if (reference.junctionTable) {
+        computedReference.junctionTable = _.camelCase(reference.junctionTable);
+      }
 
       if (reference.targetKey) {
-        const expectedConventionalTargetKeyName = underscored
-          ? _.snakeCase(reference.targetKey) : _.camelCase(reference.targetKey);
-        const targetKeyColumnUnconventional = reference.targetKey
-          !== expectedConventionalTargetKeyName;
-        return {
-          ...reference,
-          isBelongsToMany,
-          targetKeyColumnUnconventional,
-        };
+        computedReference.targetKeyColumnUnconventional = reference.targetKey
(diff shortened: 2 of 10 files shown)


arnaudbesnier:
Task linked: [CU-4embh7 P1 - Schema Generation - Table with underscores generated twice and with timestamps fields](https://app.clickup.com/t/4embh7)

services/dumper.js
```diff
@@ -200,23 +200,21 @@ function Dumper(config) {
...
-          ? _.snakeCase(reference.targetKey) : _.camelCase(reference.targetKey);
-        const targetKeyColumnUnconventional = reference.targetKey
-          !== expectedConventionalTargetKeyName;
-        return {
-          ...reference,
-          isBelongsToMany,
-          targetKeyColumnUnconventional,
-        };
+        computedReference.targetKeyColumnUnconventional = reference.targetKey
+          !== (underscored ? _.snakeCase(reference.targetKey) : _.camelCase(reference.targetKey));
```
arnaudbesnier, line 214: The previous code was already hard to read.
Please don't "compact" it for the sake of optimisation, "beauty",...

test-fixtures/mssql/sample_table.sql
```diff
@@ -0,0 +1,4 @@
+CREATE TABLE [dbo].sample_table (
+  id INT NOT NULL,
+  PRIMARY KEY (id)
+);
```
arnaudbesnier, line 4: I am not so sure this samble table is necessary for your "underscorde_no_fields" table.
You can pick any other existing table for you foreign key, it will avoid to add to much code.

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -371,6 +371,12 @@ async function analyzeSequelizeTables(databaseConnection, config, allowWarning)
   const referencesPerTable = createAllReferences(databaseSchema, schemaAllTables);
   Object.keys(referencesPerTable).forEach((tableName) => {
     schemaAllTables[tableName].references = _.sortBy(referencesPerTable[tableName], 'association');
+
+    if (!schemaAllTables[tableName].fields.length) {
+      schemaAllTables[tableName].options.underscored = isUnderscored(
+        schemaAllTables[tableName].references.map(({ foreignKey }) => ({ nameColumn: foreignKey })),
+      );
```
arnaudbesnier, line 378: Not sure to understand the logic here.
I wonder if this is another fix (vs the `computedReference.junctionTable` in the file below)

test-fixtures/mssql/sample_table.sql
closingin: Yeah that's what I did at first, but adding a new references to these tables for my specific test breaks all others, as test checks are based on raw diffs with an expected file. I felt like it was better than fixing a working test.

Tell me.

services/dumper.js
closingin: That was mostly for performance reasons (unnecessary variable creation), but I'll rollback.

services/analyzer/sequelize-tables-analyzer.js
closingin: Technically this is another fix, but both are related to the associated ClickUp task.

When a table contains no fields, `options.underscored` will always be `false` because it's computed in the fields analysis process (before references checks). That doesn't mean that the table is not snake_cased though, as it's possible that there are FKs that  are actually snake_cased.

The ideal solution would be to loop through schemaAllTables only when fields AND references are analyzed, but that would lead to a performance impact.

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -371,6 +371,12 @@ async function analyzeSequelizeTables(databaseConnection, config, allowWarning)
   const referencesPerTable = createAllReferences(databaseSchema, schemaAllTables);
   Object.keys(referencesPerTable).forEach((tableName) => {
     schemaAllTables[tableName].references = _.sortBy(referencesPerTable[tableName], 'association');
+
+    if (!schemaAllTables[tableName].fields.length) {
```
larcin, line 375: [MISC] You could add a comment here explaining why you're doing this

services/dumper.js
larcin: I do agree with Arnaud, variable creation is an acceptable performance drop to gain in code readability

test/services/database-analyzer/database-analyzer-sequelize.test.js
```diff
@@ -64,5 +64,16 @@ describe('services > database analyser > Sequelize', () => {
       expect(result.rentals).toStrictEqual(expected.rentals);
       await sequelizeHelper.close();
     }, 10000);
+
+    it('should detect snake_case even with no fields in the table', async () => {
```
larcin, line 68: [TESTS] This test only covers one of the two issues. Could you add one that checks for the other one ?

services/analyzer/sequelize-tables-analyzer.js
closingin: Fixed

services/dumper.js
closingin: Fixed

test/services/database-analyzer/database-analyzer-sequelize.test.js
closingin: The other one is already handled by the diff here : https://github.com/ForestAdmin/lumber/pull/445/commits/dee97a341c778821ebae6a993f3e6284e740bec8#diff-f40ec2acf55e196d0bce15c63549422cL28

The previous test was erroneous as it tested the wrong pattern (snake_case in `through`), when it should have been camelCase. My fix made this test stop passing, renaming the value to camelCase is the actual test.

closingin:
> I'm not sure it's related but tasks look to be displayed in a bad way. Could you try it on your end?

@larcin Pulling forestadmin-server made it work as it used to, it's ok for me!

Issue #439, opened 2020-06-15
DATABASE_URL hard-coded in docker-compose.yml


albertinator:
## Expected behavior

Expect `services.app` in `docker-compose.yml` to source environment variables from `.env`.

## Actual behavior

It was surprising to see that our `DATABASE_URL` was directly added to the `services.app.environment` property (hard-coded) when the `docker-compose.yml` was generated by `lumber`.

## Failure Logs

Not a failure, just a security concern because many users likely will create a Git repo out of this and the `docker-compose.yml` would part of that codebase with a `DATABASE_URL` hard-coded in.

I wouldn't have known this if I didn't thoroughly inspect every file generated by `lumber`.

## Context

* Lumber Package Version: 3.6.0
* Database Dialect: PostgreSQL
* Database Version: 11.6

rap2hpoutre:
A new version of lumber has just been released thanks to your suggestion: https://github.com/ForestAdmin/lumber/pull/464#issuecomment-709057240 🎉

Thank you for your feedback! 🙏 

Feel free to re-open this issue if you have any issue!

PR #464, opened 2020-09-08
feat(security): use env variables in docker-compose.yml

rap2hpoutre:
It seems it works. **Not tested on Linux though**. See: https://github.com/ForestAdmin/lumber/issues/439

## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/services/dumper.js
+++ b/services/dumper.js
@@ -305,6 +305,7 @@ function Dumper(config) {
 
   function writeDockerCompose() {
     const databaseUrl = `\${${isLinuxBasedOs() ? 'DATABASE_URL' : 'DOCKER_DATABASE_URL'}}`;
+    const forestUrl = process.env.FOREST_URL ? `\${FOREST_URL-${process.env.FOREST_URL}}` : false;
     copyHandleBarsTemplate({
       source: 'app/docker-compose.hbs',
       target: 'docker-compose.yml',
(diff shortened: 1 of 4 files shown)


guillaumedeslandes:
First comment before testing: shouldn't both `APPLICATION_PORT` and `DATABASE_SSL` be also read from the environment? To ease update for the user.

`NODE_TLS_REJECT_UNAUTHORIZED` and `FOREST_URL` are supposed to never change right?

Do not know if `APPLICATION_PORT` can used in the `ports:` section, but it would seem logical to me, for the "external" port at least.

rap2hpoutre:
You are right: `docker-compose` should not contain generated values anymore: it should entirely rely on `.env` file so users understand where to update their conf.  Let me update this PR!

rap2hpoutre:
Not sure why I can't make `docker-compose up` properly work on a fresh install on my computer (master or this branch) still it seems the env variable replacement works. Could you try to test it @guillaumedeslandes ?

guillaumedeslandes:
In `Dockerfile.hbs` there is a `EXPOSE {{ port }}`. I thing it should take its value at runtime like others in `docker-compose.hbs` via environment variables.

guillaumedeslandes (approved):

PR #471, opened 2020-09-17
feat: add dialectOptions to allow users to choose their connection options

rap2hpoutre:
An alternative version of https://github.com/ForestAdmin/lumber/pull/468

See:
- https://github.com/ForestAdmin/lumber/issues/452
- https://forest.slite.com/app/channels/-KdBU14ToWBAerCk_HGS/notes/q24dZlQGGK
- https://app.clickup.com/t/82nzna
- https://community.forestadmin.com/t/ssl-issues-with-forestadmin/837
- https://forestadmin.productboard.com/roadmap/1782065-roadmap-forest-by-lanes/features/5553384/insights
- https://forestadmin.slack.com/archives/GCYMMU085/p1599721557051000

## Pull Request checklist:

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Create automatic tests
- [x] No automatic tests failures
- [x] Test manually the implemented changes
- [x] Review my own code (indentation, syntax, style, simplicity, readability)
- [x] Wonder if you can improve the existing code

PR #468, opened 2020-09-15
feat: add dialectOptions to allow users to choose their connection options

rap2hpoutre:
See:
- https://github.com/ForestAdmin/lumber/issues/452
- https://forest.slite.com/app/channels/-KdBU14ToWBAerCk_HGS/notes/q24dZlQGGK
- https://app.clickup.com/t/82nzna
- https://community.forestadmin.com/t/ssl-issues-with-forestadmin/837
- https://forestadmin.productboard.com/roadmap/1782065-roadmap-forest-by-lanes/features/5553384/insights
- https://forestadmin.slack.com/archives/GCYMMU085/p1599721557051000

## ⚠️ Issue
I see one big issue with this implementation. Users could miss that a `dialectOptions` option has been configured since it is only referenced in `models/index.js`. For instance, if they selected `postgreSQL` during their onboarding (see specs), it will automatically add the line above in their code, and they could forget to remove it when going to production, which is a security issue since they are not protected about _man in the middle_ attack:

```js
dialectOptions: {"rejectUnauthorized":false},
```

Reviewer, before testing and reading code, could you help me decide how to address this issue?

### Solution 1

Adding a comment such as: 

```js
// You will have to review dialect options when switching environment.
// For instance, if you set `rejectUnauthorized` to false in development mode, 
// you should remove it in production to avoid security issues.
dialectOptions: {"rejectUnauthorized":false},
```

**Caveat**: User can still miss this comment (but Forest Admin has done its job: users may be aware thanks to this comment).

### Solution 2

Move `dialectOptions` JSON string in `.env`.

```env
DATABASE_DIALECT_OPTIONS={"rejectUnauthorized":false}
```

Then in `models/index.js`

```js
dialectOptions: JSON.parse(process.env.DATABASE_DIALECT_OPTIONS || '{}'),
```

**Caveat**: writing JSON in `.env` is not that common I guess.

## Pull Request checklist:

- [x] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Create automatic tests
- [x] No automatic tests failures
- [ ] Test manually the implemented changes
- [x] Review my own code (indentation, syntax, style, simplicity, readability)
- [x] Wonder if you can improve the existing code

Issue #452, opened 2020-07-18
Lumber not allowing pass-through of dialectOptions for Sequelize


antoniojtorres:
This may be more of a feature request / limitation than a bug. Thanks in advance.

## Expected behavior

Setting up ForestAdmin using lumber for the first time. Have a backend service in heroku using Heroku Postgres. When generating the lumber command and adding my database connection. Once that runs I expect lumber to run successfully.

## Actual behavior

When running the command  I received the error "SequelizeConnectionError: self signed certificate". This is because heroku postgres uses self-signed certificates for this database. 

When specifying ssl as true, lumber-generate.js generates these options for Sequelize
{ logging: false, dialectOptions: { ssl: true } }

This isn't enough for my scenario.

The installation can be pushed through with
{ logging: false, dialectOptions: { ssl: true, rejectUnauthorized: false } }

This issue is not about what is right or wrong in terms on how to connect. What I am suggesting is providing the ability to have a pass-through for database specific flags so users can tackle their edge cases. The current implementation in lumber-generate has a very narrow set of options.

## Failure Logs

? Does your database require a SSL connection?  Yes
✖ Connecting to your database
> Cannot connect to the database due to the following error:
> SequelizeConnectionError: self signed certificate

## Context

Node 12.16.3 
NPM 6.14.4

* Lumber Package Version: 3.6.4
* Database Dialect: PostgreSQL 
* Database Version: 12.3

Thank you all for the wonderful code and product. Big fan.

PS: As you may have figured, when I launched the admin the cert validation error is also present there. I am stuck :(

rap2hpoutre:
Hi @antoniojtorres! Thank you for your feedback and this detailed explanation 👍 . I guess we should improve this, I opened an issue on our product board about this need.

snwfdhmp:
Hi, I need this too !

djpate:
Same, I can't use DigitalOcean managed DB because of this.

snwfdhmp:
I cannot use Heroku managed DB myself because of this.

rap2hpoutre:
@snwfdhmp & @djpate Thank you for your feedback. This concern has been reported to the product team which is aware of the issue. This is currently in our roadmap. You will be notified as soon as it has been released!

snwfdhmp:
Thanks a lot ! You can test if this works with Heroku simply via their free Postgres add-ons

rap2hpoutre:
[...]

PR #472, opened 2020-09-18
feat: trust SSL certificate by default to ease users onboarding

rap2hpoutre:
As suggested by @arnaudbesnier, we do not add a `dialectOptions` option. See two dead PRs here: #471 & #468 

Let's add `DATABASE_REJECT_UNAUTHORIZED=false` every time, for each case. It solves P1 and it doesn't try to solve problems that don't exist.

See: 
- https://github.com/ForestAdmin/lumber/issues/452
- https://app.clickup.com/t/82nzna
- https://community.forestadmin.com/t/ssl-issues-with-forestadmin/837
- https://forestadmin.productboard.com/roadmap/1782065-roadmap-forest-by-lanes/features/5553384/insights
- https://forestadmin.slack.com/archives/GCYMMU085/p1599721557051000

## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/services/database.js
+++ b/services/database.js
@@ -57,7 +57,6 @@ function Database() {
 
   this.connect = (options) => {
     const isSSL = options.dbSSL || options.ssl;
-    let connection;
     const databaseDialect = getDialect(options.dbConnectionUrl, options.dbDialect);
 
     if (databaseDialect === 'mongodb') {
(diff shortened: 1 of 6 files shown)


services/database.js
```diff
@@ -66,30 +65,15 @@ function Database() {
...
-    } else {
-      connectionOptionsSequelize.dialectOptions = {
-        ssl: isSSL,
-      };
+    if (databaseDialect === 'mssql') {
+      connectionOptionsSequelize.dialectOptions = { options: { encrypt: isSSL } };
+    } else if (isSSL) {
+      // Add SSL options only if the user selected SSL mode.
+      // SSL Cerificate is always trusted during `lumber generate` command to ease their onboarding.
+      connectionOptionsSequelize.dialectOptions = { ssl: { rejectUnauthorized: false } };
```
jeffladiray, line 73: [Question] Shouldn't this be done also for `mssql` ? I'm not aware of this being an issue, but just want to be sure

templates/app/env.hbs
```diff
@@ -7,6 +7,8 @@ DATABASE_URL={{ databaseUrl }}
 DATABASE_SCHEMA={{ dbSchema }}
 {{/if}}
 DATABASE_SSL={{ ssl }}
+# This should be removed in production environment.
+DATABASE_REJECT_UNAUTHORIZED=false
```
jeffladiray, line 11: [MISC] Shouldn't this env var be added only when `DATABASE_SSL` is `true`?
As discussed, not a big fan of `REJECT_UNAUTHORIZED=false` by default (Especially when a user ask specifically for SSL), but I guess this has been discussed

jeffladiray (approved):
I've also attached the 2 comments I made you in private, just to keep a track on this.

Otherwise, it seems to fix the original Heroku/self-signed certificate issue, so SGTM :tada:

services/database.js
rap2hpoutre: `mssql` was already considered different and doesn't have the same options. However, the default behavior is already to trust certificates. Source: http://tediousjs.github.io/tedious/api-connection.html

> options.trustServerCertificate
If "true", the SQL Server SSL certificate is automatically trusted when the communication layer is encrypted using SSL. If "false", the SQL Server validates the server SSL certificate. If the server certificate validation fails, the driver raises an error and terminates the connection. Make sure the value passed to serverName exactly matches the Common Name (CN) or DNS name in the Subject Alternate Name in the server certificate for an SSL connection to succeed. (default: true).

templates/app/env.hbs
rap2hpoutre: > Shouldn't this env var be added only when DATABASE_SSL is true?

You may be right. Still, assuming users could switch `DATABASE_SSL` to `true/false`, they also should be able to change this options (or at least to be aware it exists). 

> As discussed, not a big fan of REJECT_UNAUTHORIZED=false by default (Especially when a user ask specifically for SSL), but I guess this has been discussed

Yes, it has been done by design. It is in `.env` so the developer will not miss it when creating a new environment (eg: prod, staging). So I guess it's OK.

PR #482, opened 2020-10-28
improvement: delegate the handling of cors request for forest routes to forest-express

ghusse:
## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/templates/app/app.hbs
+++ b/templates/app/app.hbs
@@ -26,7 +26,8 @@ if (process.env.CORS_ORIGINS) {
   allowedOrigins = allowedOrigins.concat(process.env.CORS_ORIGINS.split(','));
 }
 
-app.use(cors({
+// Let forest-express handle cors for routes starting by /forest/
+app.use(/^(?!.*\/forest\/).*$/, cors({
   origin: allowedOrigins,
   allowedHeaders: ['Authorization', 'X-Requested-With', 'Content-Type'],
   maxAge: 86400, // NOTICE: 1 day


templates/app/app.hbs
```diff
@@ -26,7 +26,8 @@ if (process.env.CORS_ORIGINS) {
   allowedOrigins = allowedOrigins.concat(process.env.CORS_ORIGINS.split(','));
 }
 
-app.use(cors({
+// Let forest-express handle cors for routes starting by /forest/
+app.use(/^(?!.*\/forest\/).*$/, cors({
```
arnaudbesnier, line 30: It is not completely clear to me which urls are supposed to be bypassed here.
Are my tests incorrect or is there a mistake in the regex?
Screenshot 2020-10-28 at 14 40 14

arnaudbesnier: Which Forest route does not start with `/forest/`?

ghusse: @arnaudbesnier All urls starting by `/forest/` should not match the regex. That's what I'm seeing in your test.

The thing is, that in `forest-express` we also have some code to handle cors for routes starting by `/forest/`, but this code is never called because requests are already handled by the declaration that I changed.

arnaudbesnier: yes `forest-express` CORS middleware is useful if you do an "in-app" integration.
It enables not to have to declare CORS manually.

As with lumber we prepare the backend to support controller overwrite, we are forced to declare the CORS manually here. If you don't do that, I am quite sure that Smart Action routes or overwritten routes will be refused because of CORS. A few tests could say if I am right or wrong.

Another way to see things, is that, adding the `/forest/*` routes exclusion you'll make this middleware useless, so it will have the same effect to remove it entirely. It is another easy test that can be done to show where it could break things.
One thing is sure is we added this seconds CORS middleware for a reason and I don't think the reason is obsolete.

ghusse: You're right, it does not work well with some routes. I don't see how to let `forest-express` handle cors for authentication routes.

I'll have to add an exception for the origin `'null'` for all routes :-/

PR #484, opened 2020-10-30
feat: allow calls from redirections

ghusse:
When an AJAX call is redirected to another endpoint, the origin is set to the string `'null'` instead of the original origin. Servers have to explicitly allow this value and return `null` in the corresponding cors header.

As we are preparing the authentication flow with OIDC, we will need to allow this origin, at least for authentication routes.

I did not find an easy way to let `forest-express` handle these cors declarations and only accept this value for routes that needed it. So the simpler solution is to allow `null` for all routes.

## Pull Request checklist:

- [ ] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [ ] Create automatic tests

### DIFF
--- a/templates/app/app.hbs
+++ b/templates/app/app.hbs
@@ -20,7 +20,7 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'public')));
 
-let allowedOrigins = [/\.forestadmin\.com$/{{#if forestUrl }}, /localhost:\d{4}$/{{/if}}];
+let allowedOrigins = [/\.forestadmin\.com$/, 'null'{{#if forestUrl }}, /localhost:\d{4}$/{{/if}}];
 
 if (process.env.CORS_ORIGINS) {
   allowedOrigins = allowedOrigins.concat(process.env.CORS_ORIGINS.split(','));
--- a/templates/app/app.hbs
+++ b/templates/app/app.hbs
@@ -20,18 +20,25 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'public')));
 
-let allowedOrigins = [/\.forestadmin\.com$/, 'null'{{#if forestUrl }}, /localhost:\d{4}$/{{/if}}];
+let allowedOrigins = [/\.forestadmin\.com$/{{#if forestUrl }}, /localhost:\d{4}$/{{/if}}];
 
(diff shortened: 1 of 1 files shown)


templates/app/app.hbs
```diff
@@ -20,7 +20,7 @@ app.use(bodyParser.urlencoded({ extended: false }));
 app.use(cookieParser());
 app.use(express.static(path.join(__dirname, 'public')));
 
-let allowedOrigins = [/\.forestadmin\.com$/{{#if forestUrl }}, /localhost:\d{4}$/{{/if}}];
```
arnaudbesnier, line 23: Silly question.
Couldn't we keep the original `allowedOrigins` for all routes except sessions ones.
And do something like this for sessions routes:
```javascript
const originsForSessions = allowedOrigins.concat('null');
app.use('/forest/sessions', cors({
  origin: originsForSessions,
  allowedHeaders: ['Authorization', 'X-Requested-With', 'Content-Type']
  maxAge: 86400, // NOTICE: 1 day
  credentials: true,
}))

app.use(cors({ /* ... */ }));
```

ghusse: Yes we can.

ghusse: The new routes for oidc are under `/forest/authentication`

templates/app/app.hbs
```diff
@@ -26,13 +26,20 @@ if (process.env.CORS_ORIGINS) {
...
+const corsConfig = {
   origin: allowedOrigins,
   allowedHeaders: ['Authorization', 'X-Requested-With', 'Content-Type'],
   maxAge: 86400, // NOTICE: 1 day
   credentials: true,
+};
+
+app.use('/forest/authentication', cors({
+  ...corsConfig,
+  origin: corsConfig.origin.concat('null')
```
arnaudbesnier, line 38: Looks great to me if it fits your technical needs for the new authentication system implementation.

One remaining questions, I have in mind is: as this generated code will be in the hands of developers, shouldn't we add a comment about this "strange" `'null'` origin? Maybe to say that it is necessary for the OpenId authentication protocol. It could prevent some developers to get rid of it, as the code looks a bit complex for no obvious reason.

arnaudbesnier (approved):
👍 
I wonder if we shouldn't add a comment to make the code a bit explicit.

templates/app/app.hbs
ghusse: Ok, I'm adding a comment

PR #486, opened 2020-11-05
chore(cors): modify cors middleware order for avoiding json parsing issue with xhr

GuillaumeCisco:
## Pull Request checklist:

- [ ] Create automatic tests

Fix: https://github.com/ForestAdmin/forestadmin/pull/2614

### DIFF
--- a/templates/app/app.hbs
+++ b/templates/app/app.hbs
@@ -28,12 +28,6 @@ const corsConfig = {
 };
 
 app.use(morgan('tiny'));
-app.use(cors(corsConfig));
-app.use(bodyParser.json());
-app.use(bodyParser.urlencoded({ extended: false }));
-app.use(cookieParser());
-app.use(express.static(path.join(__dirname, 'public')));
-
 app.use('/forest/authentication', cors({
   ...corsConfig,
   // The null origin is sent by browsers for redirected AJAX calls
@@ -41,6 +35,11 @@ app.use('/forest/authentication', cors({
   // redirects to the callback route
   origin: corsConfig.origin.concat('null')
 }));
+app.use(cors(corsConfig));
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(cookieParser());
+app.use(express.static(path.join(__dirname, 'public')));
 
 app.use(jwt({
   secret: process.env.FOREST_AUTH_SECRET,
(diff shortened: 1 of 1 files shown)


templates/app/app.hbs
```diff
@@ -33,6 +27,13 @@ const corsConfig = {
...
 };
 
+app.use(morgan('tiny'));
+app.use(cors(corsConfig));
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: false }));
+app.use(cookieParser());
+app.use(express.static(path.join(__dirname, 'public')));
+
 app.use('/forest/authentication', cors({
```
arnaudbesnier, line 37: Unfortunately, this route must be declared before:
`app.use(cors(corsConfig))`

arnaudbesnier (changes requested):

templates/app/app.hbs
GuillaumeCisco: Thank you. I will update, the only thing we need is to have cors before body-parser :)

guillaumedeslandes: Any hint why @arnaudbesnier ? On the server, `app.use(cookieParser());` is after `app.use(dynamicCors);`.

arnaudbesnier: @ghusse is the best person to explain why:
`app.use('/forest/authentication', cors({...});`
is before
`app.use(cors(corsConfig));`

See the recent contribution about it:
https://github.com/ForestAdmin/lumber/pull/484

I don't get what you have in mind about your last sentence:
> On the server, app.use(cookieParser()); is after app.use(dynamicCors);

slimee: maybe set @ghusse a reviewer of this?

ghusse: In my PR, `app.use('/forest/authentication', cors({...});` was before `app.use(cors(corsConfig));` but after bodyparser (it was the same as before my modification).

I don't know why the order changed in the meantime.

Anyway, I don't see any problem with calling `cors` before `bodyParser`. The code inside `cors` only needs to access to headers.

ghusse:
👍  it's OK for me

arnaudbesnier (approved):
The code seems nice like that.
Not sure about the commit message.
Looks like a fix to me. If so, the message should describe the potential issue for users.

PR #487, opened 2020-11-05
test: avoid dropAllCollections error when having system tables

rap2hpoutre:
Running the tests _twice_ could lead to this obscure error when `system` tables has been created:

```
MongoError: ns not found
```

In this [PR](https://github.com/ForestAdmin/lumber/pull/485), I created a test on a view: a view automatically creates a system table. This table cannot be dropped, so it has to be ignored (it crashes when you run test twice if you do not ignore it). With this fix, you can run the test ∞ times.

## Pull Request checklist:

- [ ] Create automatic tests

### DIFF
--- a/test-utils/mongo-helper.js
+++ b/test-utils/mongo-helper.js
@@ -37,12 +37,11 @@ class MongoHelper {
 
   async dropAllCollections() {
     const collections = await this.db.listCollections().toArray();
-    return Promise.all(collections.map(({ name }) => {
+    return Promise.all(collections
       // System collections are not droppable…
-      if (name.startsWith('system.')) return null;
+      .filter(({ name }) => name.startsWith('system.'))
       // …other collections are.
-      return this.db.collection(name).drop();
-    }));
+      .map(({ name }) => this.db.collection(name).drop()));
   }
 }
 
--- a/test-utils/mongo-helper.js
+++ b/test-utils/mongo-helper.js
@@ -37,7 +37,12 @@ class MongoHelper {
 
   async dropAllCollections() {
     const collections = await this.db.listCollections().toArray();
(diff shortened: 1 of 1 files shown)


test-utils/mongo-helper.js
```diff
@@ -37,7 +37,12 @@ class MongoHelper {
 
   async dropAllCollections() {
     const collections = await this.db.listCollections().toArray();
-    return Promise.all(collections.map(({ name }) => this.db.collection(name).drop()));
+    return Promise.all(collections.map(({ name }) => {
+      // System collections are not droppable…
+      if (name.startsWith('system.')) return null;
+      // …other collections are.
+      return this.db.collection(name).drop();
+    }));
```
slimee, line 45: what about 
```
collections
   .filter(({}) => !name.startsWith('system.'))
   .map(...)
```

slimee: Did you saw an issue fixed by this, like a strange behavior because system objects were dropped?

slimee (changes requested):
👍 👏

test-utils/mongo-helper.js
rap2hpoutre: > Did you saw an issue fixed by this, like a strange behavior because system objects were dropped?

Yes, see description above: https://github.com/ForestAdmin/lumber/pull/487#issue-516178015

rap2hpoutre: > what about
> 
> ```
> collections
>    .filter(({}) => !name.startsWith('system.'))
>    .map(...)
> ```

Way better! Thanks 🙏 (fixed)

slimee (approved):
🚀

PR #494, opened 2020-11-19
feat(authentication): allow to login using OpenId, without entering a password

ghusse:
## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/services/oidc/authenticator.js
+++ b/services/oidc/authenticator.js
@@ -11,8 +11,10 @@ class OidcAuthenticator {
     this.env = context.env;
     /** @private @readonly */
     this.process = context.process;
+    /** @private @readonly */
+    this.open = context.open;
 
-    ['openIdClient', 'env', 'process'].forEach((name) => {
+    ['openIdClient', 'env', 'process', 'open'].forEach((name) => {
       if (!this[name]) throw new Error(`Missing dependency ${name}`);
     });
   }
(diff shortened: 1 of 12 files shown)


services/oidc/authenticator.js
```diff
@@ -0,0 +1,104 @@
...
+  /**
+   * @private
+   * @param {import('openid-client').DeviceFlowHandle} flow
+   * @returns {Promise<import('openid-client').TokenSet>}
+   */
+  async waitForAuthentication(flow) {
+    const expiresIn = flow.expires_in;
+    try {
+      this.process.stdout.write(`Navigate to the url: ${flow.verification_uri}\n`);
+      this.process.stdout.write(`and enter your verification code: ${flow.user_code}\n`);
```
DrRaider, line 71: Why not use the logger here ?

DrRaider, line 71: Why not use the logger here ?

DrRaider (changes requested):

services/oidc/authenticator.js
ghusse: Because that does not correspond to logs, they are some messages for the user. Logs are appended to the console with the current configuration, but that's not necessarily always the case (depending on the config).

Here, I want to always display some content to the user.

I hesitated, but this way seems more clean. That's also why I did not use `console.log`, because it corresponds more to a stdout output even if the result is the same.

DrRaider (approved):

PR #507, opened 2020-12-01
refactor: switch dumper to class

arnaud-moncel:
## Pull Request checklist:

- [X] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [x] Create automatic tests
- [X] No automatic tests failures
- [X] Review my own code (indentation, syntax, style, simplicity, readability)

### DIFF
--- a/test/services/dumper/dumper.test.js
+++ b/test/services/dumper/dumper.test.js
@@ -30,7 +30,7 @@ async function createLinuxDump(overrides = {}) {
     ...overrides,
   };
 
-  const dumper = await new Dumper(config);
+  const dumper = new Dumper(config);
   await dumper.dump({});
 }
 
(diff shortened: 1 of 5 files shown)


test/services/dumper/dumper.test.js
```diff
@@ -163,4 +163,43 @@ describe('services > dumper', () => {
       });
     });
   });
+
+  describe('getDatabaseUrl', () => {
+    it('should return the connextion string if no dbConnectionUrl is provided', () => {
```
jeffladiray, line 168:
```suggestion
    it('should return the connection string if no dbConnectionUrl is provided', () => {
```

jeffladiray (commented):
I only found a typo in a test label. Otherwise, LGTM.

Maybe it would be best to also switch it to dependency injection, or you prefer to do it in a following PR ?

Also, i'm not sure about the `chore` in the commit message, maybe `refactor` would suit better here?

jeffladiray (approved):
LGTM 👍

PR #511, opened 2020-12-10
refactor: remove globally defined sequelize variables from sequelize-tables-analyzer

arnaud-moncel:
## Pull Request checklist:

- [X] Write an explicit title for the Pull Request, following [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/#summary)
- [ ] Create automatic tests
- [X] No automatic tests failures
- [X] Review my own code (indentation, syntax, style, simplicity, readability)

### DIFF
--- a/services/analyzer/sequelize-tables-analyzer.js
+++ b/services/analyzer/sequelize-tables-analyzer.js
@@ -15,16 +15,12 @@ const ASSOCIATION_TYPE_HAS_ONE = 'hasOne';
 
 const FOREIGN_KEY = 'FOREIGN KEY';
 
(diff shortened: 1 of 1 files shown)


jeffladiray (approved):
LGTM.
It wasn't easy to review, but I tested manually and check the coverage for lines that wouldn't be covered and everything seems fine on my end.

The commit message is a bit generic though, + lumber generate does not support onboarding on multiple database, so I would suggest a more "technical" message, maybe something like
`refactor: remove globally defined sequelize variables from sequelize-tables-analyzer`? WDYT?

PR #512, opened 2020-12-10
refactor(database): switched database to class with injection

jeffladiray:
## Pull Request checklist:

- [x] Create automatic tests

### DIFF
--- a/test/services/database.unit.test.js
+++ b/test/services/database.unit.test.js
@@ -131,6 +131,7 @@ describe('services > database', () => {
           },
         },
       };
+
       describe('when connectionUrl is not provided', () => {
         it('should build the connection URL and call mongodb connect', async () => {
           expect.assertions(1);
--- a/test/services/database.unit.test.js
+++ b/test/services/database.unit.test.js
@@ -1,0 +1,352 @@
+const Database = require('../../services/database');
+
+const setupDatabase = ({ Sequelize = {}, mongodb = {}, terminator = {} } = {}) =>
+  new Database({
+    Sequelize,
+    mongodb,
+    terminator,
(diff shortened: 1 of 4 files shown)


test/services/database.unit.test.js
```diff
@@ -0,0 +1,352 @@
...
+        .mockImplementation(jest.fn);
+
+      await database.connectToMongodb({});
+
+      expect(handleAuthenticationErrorSpy).toHaveBeenCalledTimes(1);
+      expect(handleAuthenticationErrorSpy).toHaveBeenCalledWith(connectError);
+    });
+  });
+
+  describe('connect', () => {
```
arnaud-moncel, line 251: What about ssl tests? And tests of getDialect call?

arnaud-moncel (commented):
Nice to see a lot of tests ❤️ 
I have two comments about tests feel free to discuss them.

test/services/database.unit.test.js
jeffladiray: >  And tests of getDialect call?

You mean in this function, right? Not sure that's mandatory (Since I already check for the createMongoose/sequelize connection)

> ssl tests?

Requires 3 tests to be fullfilled, which is about 75 LOC. Would prefer not to write these, since (to me at least) it indicate that some code should be refactored. I would prefer to refactor the code here in a following PR instead of adding more tests (If that's ok to you).

arnaud-moncel: ok for me ;)

arnaud-moncel (approved):
🚀

arnaud-moncel (approved):
LGTM

PR #540, opened 2021-02-15
chore(ci): move from travis to github action

arnaudbesnier:
## Pull Request checklist:

- [ ] Create automatic tests


.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: yarn lint
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    needs: [lint]
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    strategy:
+      matrix:
+        node: ['10', '14']
```
arnaudbesnier, line 40: Didn't find a way to express 'LTS', after a quick search.

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Login on dockerhub
+        run: echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
+      - name: Start docker container
+        run: docker-compose up -d; sleep 15
+      - name: Send coverage
+        uses: paambaati/codeclimate-action@v2.7.4
+        env:
+          CC_TEST_REPORTER_ID: ${{ secrets.CC_TEST_REPORTER_ID }}
+        with:
+          coverageCommand: yarn test:coverage
```
arnaudbesnier, line 62: It will send the coverage twice, it was not the case with Travis.
It may be a problem.

.github/workflows/build.yml
VincentMolinie: It's more than enough I think Once there will be w new LTS we will add it to the matrix

.github/workflows/build.yml
```diff
@@ -0,0 +1,89 @@
...
+    runs-on: ubuntu-latest
+    needs: [lint]
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
+    strategy:
+      matrix:
+        node: ['10', '14']
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
VincentMolinie, line 44: This is only needed for the commit-lint

.github/workflows/build.yml
VincentMolinie: Well may be we should send the coverage only if we are using node 14 and just laucnh the test in the other cases

VincentMolinie (changes requested):

.github/workflows/build.yml
```diff
@@ -54,8 +52,12 @@ jobs:
         run: echo ${{ secrets.DOCKER_PASSWORD }} | docker login -u ${{ secrets.DOCKER_USERNAME }} --password-stdin
       - name: Start docker container
         run: docker-compose up -d; sleep 15
-      - name: Send coverage
+      - name: Test Node 10
+        if: matrix.node == '10'
```
VincentMolinie, line 56: I would more say if not 14 then test. Like that if we want to add node 12 it just works ^^

VincentMolinie: Or you could even add a variable into the matrix to know if you have to do a coverage or a simple test

VincentMolinie:
```yaml
strategy:
  matrix:
    node: ['10', '14']
    include:
      - node: 14
         withCoverage: ${{ true }}
```

```
  - name: Test with Node ${{ matrix.node }}
    if: !matrix.withCoverage
```

VincentMolinie (approved):
Please consider my comment before merging 🙏

PR #554, opened 2021-03-05
fix(analyzer): work around sequelize bugs on default values

romain-gilliotte:
This PR works around sequelize bugs when retrieving default values from the database.

Each commit deals with different issues causing the bug
- Commit 1: Simplify the dumper to ensure that it cannot output invalid javascript
- Commit 2: Refactor database analyser to fetch the schema early on when not provided by the user. This was previously done on the showAllTables function, and only for mssql
- Commit 3: Work around sequelize bugs to retrieve the default values from the databases
- Commit 4: Unit tests.

Note: As it differs quite a lot from one dialect of SQL to another I tested the default value extraction from the database only for Postgres.
It was previously completely untested.


arnaudbesnier:
Task linked: [CU-f54mcu P3 - Lumber timestamp generation - invalide default value to timestamp](https://app.clickup.com/t/f54mcu)

arnaudbesnier:
Task linked: [CU-8pnt1c P2 - Lumber - Lumber generates incorrect date default value](https://app.clickup.com/t/8pnt1c)

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -42,8 +42,66 @@ async function getDefaultSchema(connection, userProvidedSchema) {
...
+  return result;
+}
+
+/** Retrieve the description of the fields in a given table. */
+async function analyzeFields(queryInterface, tableName, config) {
+  // There is a bug in sequelize/dialects/mysql/query-generator#describe
+  // Sequelize queries over `[schema].[table]` instead of `[schema]`.`[table]`
+  // => Don't provide the schema when using mysql/mariadb
+  const dialect = queryInterface.sequelize.getDialect();
+  const workaroundConfig = ['mysql', 'mariadb'].includes(dialect) ? {} : { schema: config.dbSchema };
```
VincentMolinie, line 83: [MISC] SO if you are using a custom schema on mysql it won't work anymore right 🤔  ?

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -42,8 +42,66 @@ async function getDefaultSchema(connection, userProvidedSchema) {
...
+  `;
+
+  const rows = await queryInterface.sequelize.query(getDefaultsQuery, {
+    type: queryInterface.sequelize.QueryTypes.SELECT,
+    replacements: [config.dbSchema, tableName],
+  });
+
+  rows.forEach((row) => {
+    columnsByName[row.colname].defaultValue = parseSqlDefault(row.coldefault);
+  });
```
VincentMolinie, line 102: [MISC] Why using that columnsByName 🤔  ?

test-expected/sequelize/db-analysis-output/default-values.expected.js
```diff
@@ -1,99 +0,0 @@
-const Sequelize = require('sequelize');
```
VincentMolinie, line 1: [MISC] Did you commit the file in 2 times 🤔 . This file should just be marked as renamed

test-expected/sequelize/db-analysis-output/default_values.expected.js
```diff
@@ -0,0 +1,148 @@
+const Sequelize = require('sequelize');
```
VincentMolinie, line 1: [MISC] Can you tell me what change in here 🙏

test-expected/sequelize/dumper-output/default-values.expected.js
```diff
@@ -5,47 +5,85 @@ module.exports = (sequelize, DataTypes) => {
...
-    },
-    arrayDefault: {
-      type: DataTypes.ARRAY(DataTypes.INTEGER),
-      defaultValue: [],
+    boolNull: {
+      type: DataTypes.BOOLEAN,
+      field: 'bool_null',
     },
-    booleanDefault: {
+    boolCst: {
```
VincentMolinie, line 12: [WOrding] We don't do shortcut in namings. Plus keep the old naming it's just increasing the diff 😅

test-expected/sequelize/dumper-output/default-values.expected.js
```diff
@@ -5,47 +5,85 @@ module.exports = (sequelize, DataTypes) => {
...
+      defaultValue: "1983-05-27 00:00:00",
     },
-    dateExprDefault: {
+    dateExpr1: {
       type: DataTypes.DATE,
+      field: 'date_expr1',
       defaultValue: Sequelize.literal('now()'),
     },
-    enumDefault: {
-      type: DataTypes.ENUM('LEFT','RIGHT'),
```
VincentMolinie, line 28: Don't delete those tests please 🙏

test-expected/sequelize/dumper-output/default-values.expected.js
```diff
@@ -5,47 +5,85 @@ module.exports = (sequelize, DataTypes) => {
...
-      type: DataTypes.INT,
-      defaultValue: 42,
+    dateExpr3: {
+      type: DataTypes.DATE,
+      field: 'date_expr3',
+      defaultValue: Sequelize.literal('timezone(\'utc\'::text, now())'),
     },
-    numericWithStringNULLDefault: {
-      type: DataTypes.DOUBLE,
-      defaultValue: "NULL",
```
VincentMolinie, line 37: Same here

test-fixtures/postgres/default_values.sql
```diff
@@ -0,0 +1,20 @@
...
+    bool_cst BOOLEAN DEFAULT TRUE,
+    int_cst INTEGER DEFAULT 42,
+    str_null VARCHAR(25) DEFAULT NULL,
+    str_cst VARCHAR(25) DEFAULT 'constant',
+    str_expr VARCHAR(25) DEFAULT UPPER('Hello' || 'World'),
+    date_null TIMESTAMP DEFAULT NULL,
+    date_cst1 TIMESTAMP DEFAULT '2010-01-01T00:00:00Z',
+    date_cst2 TIMESTAMP DEFAULT '1983-05-27',
+    date_expr1 TIMESTAMP DEFAULT NOW(),
+    date_expr2 TIMESTAMP DEFAULT now(),
```
VincentMolinie, line 13: I don't see your fix 🤔 . You were supposed to fix CURRENT_TIMESTAMPS so it should be in the tests 😉

test-utils/sequelize-helper.js
```diff
@@ -41,10 +41,17 @@ class SequelizeHelper {
...
-    // eslint-disable-next-line import/no-dynamic-require, global-require
-    return require(expectedFilename);
+
+    try {
+      // eslint-disable-next-line import/no-dynamic-require, global-require
+      return require(`${expectedFilename}.json`);
+    } catch (e) {
+      // eslint-disable-next-line import/no-dynamic-require, global-require
+      return require(expectedFilename);
+    }
```
VincentMolinie, line 54: Why making it more complex 🤔  ?

VincentMolinie (changes requested):
As it's your first PR, I'll explain some rules we have at Forest:
- The simpler the better, meaning don't mix things. If you want to refacto fine, but do a PR just with the refacto
- Try to reduce as much thr work for the reviewer, this must be really simple for him. Here there is way too much info / line change for that fix. There is some refacto, some file changing, some renaming and ll of theme increase a lot the diff which takes times to review while I could may be have done it in a few minutes it took me hours 😉 
- Always fix only what you have in the title of your PR (So just the fix of your card), it look like you actually fixed something on the schema but I don't see the link with your PR 🤔 
- You can then of course do the ofther fixes in another PR put as I said before, the simpler the better ^^
- Wording: We don't do shortcut in naming 😉  (example: `cst => constant`)
- Always reproduce rhe issue in your test, so here you had an issue with CURRENT_TIMESTAMP, add it to the test 😉 
- And I think that's it for now 😅

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -434,16 +447,25 @@ async function analyzeSequelizeTables(databaseConnection, config, allowWarning)
     }
   }
 
+  // If dbSchema was not provided by user, get default one.
+  const myConfig = {
```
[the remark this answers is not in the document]
romain-gilliotte: Let's go with `configWithSchema` as otherwise it break the 100 columns limitations everywhere else

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -42,8 +42,66 @@ async function getDefaultSchema(connection, userProvidedSchema) {
...
+  let result;
+  if (!expression || expression === 'NULL' || expression.startsWith('NULL::')) {
+    result = null;
+  } else if (falses.includes(expression)) {
+    result = false;
+  } else if (trues.includes(expression)) {
+    result = true;
+  } else if (/^'(.*)'::jsonb?$/i.test(expression)) {
+    const [, content] = expression.match(/^'(.*)'::(.*)$/i);
+    result = JSON.parse(content);
```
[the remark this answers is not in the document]
romain-gilliotte: I don't think the JSON can be invalid, as it is parsed by the database when the CREATE TABLE query is run.
I'll check that is the case.

services/analyzer/sequelize-tables-analyzer.js
romain-gilliotte: The sequelize `describeTable` returns a hashmap { columnName: columnData, ...}
Is the naming bad?

services/analyzer/sequelize-tables-analyzer.js
romain-gilliotte: Sequelize is using `SHOW FULL COLUMNS FROM ${table};`, so it should work as long as it's the default schema.
However, when querying the `INFORMATION_SCHEMA` I need to provide the schema explicitly...

But you're right, this is not unit tested AFAIK

test-expected/sequelize/db-analysis-output/default-values.expected.js
romain-gilliotte: I mark this as resolved, can be discussed in the comment about test-expected/sequelize/db-analysis-output/default_values.expected.js

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -16,6 +16,32 @@ const ASSOCIATION_TYPE_HAS_ONE = 'hasOne';
...
+    postgres: 'SELECT current_schema() AS default_schema',
+  };
+
+  if (queries[dialect]) {
+    const rows = await connection.query(
+      queries[dialect],
+      { type: connection.QueryTypes.SELECT },
+    );
+
+    return rows.length ? rows[0].default_schema : 'public';
```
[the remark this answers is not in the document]
romain-gilliotte: It could, can you point me to the link in the doc? I can't find it

test-expected/sequelize/dumper-output/default-values.expected.js
romain-gilliotte: Responded in another comment

test-expected/sequelize/dumper-output/default-values.expected.js
romain-gilliotte: Responded in another comment

test-expected/sequelize/dumper-output/default-values.expected.js
romain-gilliotte: Responded in another comment

test-fixtures/postgres/default_values.sql
romain-gilliotte: I did that initially, but it is a bit of a pain as the output of the test is different depending the postgres version.
I think that on Postgres 9.x, CURRENT_TIMESTAMP() was just an alias to NOW(), and it is no longer the case in 12.x (maybe it was the opposite)

If you do `CREATE TABLE ... somecolumn TIMESTAMP DEFAULT CURRENT_TIMESTAMP`
And then later on:
`SELECT column_default FROM information_schema.columns WHERE .... AND column_name = 'somecolumn'`

On Postgres 9.x you get: `NOW()`
On Postgres 12.x you get: `CURRENT_TIMESTAMP`

This breaks the tests, as the same output is expected for all databases and versions, which is not feasible here.
I would need to provide 6 different expected outputs in `test-expected` for each one of the database backends which are tested, which was asking for more refactoring, and the PR was long enough

test-expected/sequelize/db-analysis-output/default_values.expected.js
romain-gilliotte: When reading how the test suite work, you'll see that the files in `test-expected/db-analysis-output` are used both as:
- Expected output (as their name imply) to check that the `database-analyzer` is doing its job. This happens in `test/services/database-analyzer/...`
- Input for all the `dumper` tests. This happens in `test/services/dumper/dumper-sequelize.test.js`

So it goes like this:
```
test-fixtures/postgres/addresses.sql
test-fixtures/mysql/addresses.sql      ---analyzer---> test-expected/sequelize/db-analysis-output/addresses.json ---dumper--> test-expected/sequelize/dumper-output/addresses.expected.js
test-fixtures/mssql/addresses.sql
```

This was done for all tests besides the one for default values: there is no corresponding SQL files for the "expected analysis output" file which was on the test-expected folder.

The extraction of default values from the database was left completely untested.
What was tested was only the second step (dumping).

That's why I deleted and completely rewrote this file (and lost some of the tests in the way): I wrote a SQL file so that I get to test the db-analyzer and needed the expected-ouput file to match it.

**As a matter of fact, the test coverage on this feature is better than before this PR, but it is still strongly lacking, as the default values extraction is still untested for mssql and mysql.**

romain-gilliotte:
First, thanks for taking so much time!
I'll try to reply here to the remarks

> * The simpler the better, meaning don't mix things. If you want to refacto fine, but do a PR just with the refacto
Sorry, I did refactor in two locations!

Location 1:
My problem was that when querying `information_schema`, I need to know the schema, even when the user is not providing it explicitly (I needed that to get the default values).
This was already the case for previous bug patches on this file:
- ` || public` at https://github.com/ForestAdmin/lumber/blob/master/services/analyzer/sequelize-tables-analyzer.js#L417
- mssql special case at https://github.com/ForestAdmin/lumber/blob/master/services/analyzer/sequelize-tables-analyzer.js#L34

I did not want to duplicate code, so I ended moving that in another location, and properly handling all cases.

Location 2:
https://github.com/ForestAdmin/lumber/blob/master/services/analyzer/sequelize-tables-analyzer.js#L306

I did not understand why primary keys get a special treatment in this case, so I moved the logic to a function, and applied it to all columns (primary or not), which makes the final output less random looking

> * Try to reduce as much thr work for the reviewer, this must be really simple for him. Here there is way too much info / line change for that fix. There is some refacto, some file changing, some renaming and ll of theme increase a lot the diff which takes times to review while I could may be have done it in a few minutes it took me hours 😉
> * Always fix only what you have in the title of your PR (So just the fix of your card), it look like you actually fixed something on the schema but I don't see the link with your PR 🤔
> * You can then of course do the ofther fixes in another PR put as I said before, the simpler the better ^^

I think the bug was not easy to fix...
Not my best choice for a first PR

It was a prerequisite to access `information_schema` in a reliable way.
Before doing the whole schema extraction, I had some (manual...) tests failing on mysql.

> * Wording: We don't do shortcut in naming 😉  (example: `cst => constant`)
I still need to fix those on the SQL file

> * Always reproduce rhe issue in your test, so here you had an issue with CURRENT_TIMESTAMP, add it to the test 😉
I responded to that in another comment.
I was also fixing default values like "timezone('utc'::text, now())" which are tested properly but CURRENT_TIMESTAMP proved more difficult.

> * And I think that's it for now 😅
It's a lot already 😅

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -51,24 +51,34 @@ function parseSqlDefault(expression) {
...
+      result = false;
+    } else if (trues.includes(expression)) {
+      result = true;
+    } else if (/^-?\d+(\.\d+)?$/.test(expression)) {
+      result = Number.parseFloat(expression);
+      if (result.toString() !== expression) {
+        throw new Error('Bignum overflowing javascript precision');
+      }
+    } else if (/^'.*'$/.test(expression)) {
+      result = expression.substr(1, expression.length - 2);
```
romain-gilliotte, line 67: Maybe this is missing removing the simple quote escaping.
Need to test what would happen.

Something like`.replace(/\\'/g, "'")`

VincentMolinie: May be to add in a test. I'll try to help you tomorrow on those tests  😉  (If I find some time 😅 )

test/services/database-analyzer/database-analyzer-sequelize.test.js
```diff
@@ -32,18 +37,36 @@ describe('services > database analyser > Sequelize', () => {
...
+
+      const defaultsValueExpected = {
+        [databaseUrls.DATABASE_URL_POSTGRESQL_MIN]: _.cloneDeep(expectedDefaultValuesPostgres),
+        [databaseUrls.DATABASE_URL_POSTGRESQL_MAX]: expectedDefaultValuesPostgres,
+        [databaseUrls.DATABASE_URL_MYSQL_MAX]: expectedDefaultValuesMysql,
+        [databaseUrls.DATABASE_URL_MSSQL_MIN]: expectedDefaultValuesMssql,
+        [databaseUrls.DATABASE_URL_MSSQL_MAX]: expectedDefaultValuesMssql,
+      };
+
+      defaultsValueExpected[databaseUrls.DATABASE_URL_POSTGRESQL_MIN]
```
VincentMolinie, line 51: A little comment just to explain why 🙏

VincentMolinie, line 51: WHy not defining this before the  test 🤔  ? It seems static to me

services/analyzer/sequelize-default-value.js
```diff
@@ -12,7 +12,7 @@ class DefaultValueExpression {
   }
 
   parse() {
-    if (this.expression === null || this.expression === undefined) {
+    if (!this.expression) {
```
VincentMolinie, line 15: This was fine before 😅 , this will transform empty string to null which we do not want 😉

VincentMolinie (commented):
Good job it's way more cleaner than before 👏 👏

services/analyzer/sequelize-default-value.js
romain-gilliotte: Oops. I did that because the linter was complaining over code complexity. I'll revert the change.

test/services/database-analyzer/database-analyzer-sequelize.test.js
romain-gilliotte: It is static, I could move it to the top of the file.
I put it there so that readers know for which test we're doing that.

services/analyzer/sequelize-tables-analyzer.js
```diff
@@ -42,73 +42,45 @@ async function getDefaultSchema(connection, userProvidedSchema) {
...
+      replacements: [config.dbSchema, tableName],
+    });
+
+    rows.forEach((row) => {
+      columnsByName[row.colname].defaultValue = row.coldefault;
+    });
+  }
+
+  Object.values(columnsByName).forEach((column) => {
+    const myColumn = column; // Workaround pedantic linter.
```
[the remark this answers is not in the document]
romain-gilliotte: Because getting defaults values is buggy in MsSQL and Postgres, but not for MySQL.
=> I only run the workaround with those databases, but still need to parse the default values for MySQL

VincentMolinie (approved):
Works like a charm 💯  !

Just two little things to improve (for another PR):
- Array: they are still written with Sequelize.literal, we could may be transform it 🤔 
- Json: The json is written as a simple json, meaning with double quotes around value and key. We Should not have any double quote (only single quote), and there should not be any quote around the key if not necessary
- String: They are surronded by double quote instead of single quote

PR #578, opened 2021-06-17
fix(mongoose): wrap with quotes invalid js identifiers in sub documents properties

slimee:
The P1 https://app.clickup.com/t/mnbt5x

Note: the frontend must also be patched to handle spaces. It's ok for hyphen, numbers, etc.

## Pull Request checklist:

- [x] Create automatic tests


handlerbars/helpers/wrap-quotes.js
```diff
@@ -0,0 +1,4 @@
+const Handlebars = require('handlebars');
+
+Handlebars.registerHelper('wq', (value1) => value1
+  .indexOf(' ') > -1 ? `'${value1}'` : value1);
```
slimee, line 4: We may have a regex here instead of check space in content

jeffladiray (commented):
As discussed, we already wrap top level fields with double quotes. I'm wondering if we shouldn't do something similar here (wrap ALL fields with double quotes), it would avoid the need of a new helper. However, the current implementation is not eslint-friendly, so if you want to keep your helper, I would suggest to do the same for top-level fields as well ?

Let me know, otherwise it works as expected.

jeffladiray (approved):
Maybe you could also use your `wq` helper also in the mongoose model file here -> https://github.com/ForestAdmin/lumber/blob/526770ea2b80a499aa1c3374cb63070c6b755ddc/templates/app/models/mongo-model.hbs#L9 ?


Otherwise, looks to work as expected
image

Your call!

jeffladiray (approved):

